PR #6, opened 2018-03-20
POWR-45 initial commit of form/app button

jookyg:


web/sites/default/config/field.field.paragraph.form_app_button.field_form_button.yml
```diff
@@ -0,0 +1,23 @@
...
+    - paragraphs.paragraphs_type.form_app_button
+  module:
+    - link
+id: paragraph.form_app_button.field_form_button
+field_name: field_form_button
+entity_type: paragraph
+bundle: form_app_button
+label: 'Form link'
+description: ''
+required: false
```
joshuami: Please make the field required. Paragraphs will only require them to fill it out if a Form/App Button is added to the step, but if it exists, it should be required.

web/sites/default/config/core.entity_view_display.paragraph.form_app_button.default.yml
```diff
@@ -0,0 +1,28 @@
...
+targetEntityType: paragraph
+bundle: form_app_button
+mode: default
+content:
+  field_form_button:
+    weight: 0
+    label: hidden
+    settings:
+      trim_length: 80
+      url_only: false
```
joshuami: There is a setting "Allow link text", please make it "Required" instead of "Optional"

PR #17, opened 2018-04-10
Update readme

joshuami:
@RichardDavies, let me know if these expanded git instructions help. Is there more you would add for now?

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ See other Lando with Pantheon commands at https://docs.devwithlando.io/tutorials
 To best work with Pantheon Multidev, we are going to keep feature branch names simple and use the master branch as our integration point that builds to the Pantheon Dev environment.
 
 1. ### Start new work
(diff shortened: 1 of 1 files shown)


joshuami:
Well shoot. I bungled the EVA getting added to this PR. I'll see if I can't fix that and resubmit with the additional changes you mentioned. Those are all great ideas.

joshuami:
@RichardDavies, good point about the git pull. I just added that as the first step with some instructions for optionally pruning unused branches. I also added some clean up steps for the build master at the end of each issue and sprint.

joshuami:
@jookyg, I think Lando pull has always worked from a home network. It only fails at the office because of firewall rules. That said, the database restore process I wrote into the README is taking a backup directly from Pantheon's dashboard and using the `lando db-import` command against it.

Notes about QA might be helpful. I don't really specify their steps. Let me think about that and I might add it as a follow up commit.

PR #18, opened 2018-04-11
Add the Entity Views Attach module for adding backreferences view field

joshuami:
This adds the EVA module and updates Group module to pull from latest stable 1.0 branch. That will update to RC2 after running `composer update drupal/group --with-dependencies` and `drush updb`. (Add the `lando` command if you do not have composer or drush installed locally.


jookyg:
See my comments in Slack and story POWR-49. This module may not be necessary.

PR #20, opened 2018-04-12
Powr 112

RichardDavies:
Adds and enables redirect module.


RichardDavies:
Out of curiosity, I'd like to know how you resolve the merge conflict in composer.lock file. I assume the hash changes when the dependencies change... how do you determine which hash to keep?

joshuami:
@RichardDavies, that is a great question. I've gone with the content hash from your commit. That is the first time I've seen the conflict when adding a module, so I'll be able to tell you in a moment if I made the correct choice. Research seems to indicate that it should change, but not why it would be a merge conflict rather than just a change.

RichardDavies:
@joshuami I was wondering too why the new hash wasn't just a change. I think it's probably because after I created this branch, it looks like you merged another branch into master containing the new EVA module. So now it doesn't know whether to keep the hash from the EVA module or from my branch.

joshuami:
Well, the good news is that it liked your hash just fine. So we are good to go. I'm going to accept the ticket now and merge our changes to test.

PR #28, opened 2018-04-19
POWR-60 homepage actions and topic block updates.

joshuami:


mike-mcdonald:
Is there any reason to change the machine name on these blocks from block_X to something meaningful in general, or here specifically?

To me that would depend on whether the view might be used for multiple blocks.... That might come up here since these views are pretty broadly named "Topics" and "Community Actions" and it may or may not be that we create more blocks underneath this as a means of organization.  You all might have covered this in an earlier meeting I didn't attend; I remember it being a possible topic of discussion at some point.  If so, apologies.

joshuami:
I suppose we could rename those views to be more specific. Something like "topic_homepage_block" for a block that is only to be used on the homepage.

Let's talk through possible naming convention updates in our sprint retro next week. If we decide on something, we can refactor as we go. This won't be the last time we touch these blocks since I didn't really apply any significant styling.

joshuami:
@RichardDavies @kxwang @jookyg @mike-mcdonald, would one of you be willing to give me a approval so we can stick with the process and get this merged.

PR #35, opened 2018-04-26
POWR-152 make city service mode and step fields optional for easier s…

joshuami:
…tub process.


joshuami:
Heh... I made the PR title too long 😀

joshuami:
It was so easy that no one approved it yet!

😁

PR #41, opened 2018-05-01
POWR 125 updates

jookyg:
- Added Chosen module and library
- Added Audience vocabulary
- Added Audience field to City Service, Bureau/office, Program
- Added Topics field to Program


jookyg:
Whaaaa...? I could swear I only saw the config changes when I ran git status before committing. I'll investigate.

PR #44, opened 2018-05-04
POWR-151 fixing the group displays and adding an audience block

joshuami:
This was not intended as a code change story... but I'm giving it a little umph.


joshuami:
Could someone give me a review. I'm passing this into QA as ready to merge.

PR #46, opened 2018-05-07
POWR-163 convert the service mode field to a taxonomy lookup with chosen

joshuami:


mike-mcdonald:
Watching some of the CirleCI stuff go was pretty neat!

PR #56, opened 2018-05-14
POWR-76 Confirm audiences homepage block display

brittanysmart:
Wanted to confirm that the audiences homepage block is displaying correctly.


joshuami:
I'm closing the pull request from the powr-76-1 branch. There were too many issues here and you now have 76-2 for us to finalize.

PR #59, opened 2018-05-16
POWR-190 Optimize CircleCI script

kxwang:
This PR is for CircleCI development purpose. Please ignore for now. Will start review after the code stabilizes.


kxwang:
It's down to 5-6 minutes now. 
1. Spinning up behat and run tests takes 1 minute. If we only need to check for one string, we can use curl to save 1 minute.
2. Tried hard to get rid of the 30 seconds wait but "drush cim" always fails. I've opened a ticket with Pantheon.
3. More than half of the time is spent on sending files from CircleCI to Pantheon.

When you merge, I'll need to watch the build on master. Thanks!

kxwang:
Pantheon ticket for "drush cim"
https://dashboard.pantheon.io/sites/5c6715db-abac-4633-ada8-1c9efe354629#support/ticket/104957

.circleci/config.yml
```diff
@@ -166,55 +94,48 @@ behat_tests: &behat_tests
...
             # git filter-branch -f --index-filter 'git rm -rf --cached --ignore-unmatch sites/default/files' --prune-empty -- afc758c6..HEAD
             git push -f pantheon $CIRCLE_BRANCH
-            # Deploy code from local env to the new multidev site
-            # terminus env:deploy portlandor.$CIRCLE_BRANCH --sync-content --note="Manual deploy to powr-118"
             # Wake up the multidev site
             terminus env:wake portlandor.$CIRCLE_BRANCH
-            # Check site status
-            terminus drush portlandor.$CIRCLE_BRANCH -- status
             # The following two drush commands will fail without wait for a few seconds
+            # sleep 40s
```
joshuami: This line says sleep 40s, but the actual command is sleep 30s?

PR #64, opened 2018-05-21
POWR-190-1 Run CIM before CR

kxwang:
This PR is trying to debug the problem with running "drush cim" on Pantheon.


joshuami:
@kxwang, can we close this one out? I feel like we merged something from 190 already. Do I have that wrong?

PR #76, opened 2018-05-24
POWR-130 Footer copyright

kxwang:


joshuami:
I'm just going to let the build to complete before I merge to make sure I didn't slip on the conflict resolution.

PR #85, opened 2018-05-29
POWR-126 Install & enable swiftype module

RichardDavies:


RichardDavies:
@joshuami Ok, this is ready for another review and I've got a good feeling about it this time :-)

web/sites/default/config/swiftype_integration.settings.yml
```diff
@@ -0,0 +1,7 @@
+swiftype_integration_install_key: K2y8KFxi3pKyFfU1-csZ
```
joshuami: @RichardDavies, is there any concern with these keys being published publicly? Are they secrets or just configuration information?

RichardDavies:
Oh, good question. Yeah, we need to keep that secret. Should I remove it from the config and push up another commit?

joshuami:
We'll need to reset those credentials as well. Anything we publish to github, even if we remove it, is essentially compromised because there is no way to know if they were scraped by tools that look for this sort of thing.

PR #92, opened 2018-06-01
POWR-86 Add the Project group type

kxwang:
Enabled the "Date Range" core module to support Start and End date/time.


web/modules/custom/portland/portland.module
```diff
@@ -11,7 +11,11 @@ const GROUP_CONTENT_TYPES = array('services', 'locations', 'buildings');
 function portland_group_insert(GroupInterface $group) {
 
     $gid = $group->id();
-    $shortname = substr(str_replace(" ","_",trim(strtolower($group->get('field_shortname_or_acronym')->value))),0,60);
+    // For Program/Project, there is no shortname field. Fall back to label (the title field)
+    $shortname = rawurlencode(substr(str_replace(" ","_",trim(strtolower($group->label))),0,60));
```
joshuami: I think we should make pathauto a dependency and use `$cleanedtitle = \Drupal::service("pathauto.alias_cleaner")->cleanString("MY STRING TO CLEAN");` rather than rawurlencode. This will ensure that any titles that contain special characters have those stripped following the rules at https://test.portland.gov/admin/config/search/path/settings.

web/modules/custom/portland/portland.module
```diff
@@ -11,7 +11,11 @@ const GROUP_CONTENT_TYPES = array('services', 'locations', 'buildings');
...
+    if(isset($group->field_shortname_or_acronym)) {
+        $shortname = rawurlencode(substr(str_replace(" ","_",trim(strtolower($group->get('field_shortname_or_acronym')->value))),0,60));
```
joshuami: Same as above.

PR #99, opened 2018-06-08
POWR-183 Added a custom module for using cleave.js for telephone widgets

mike-mcdonald:


mike-mcdonald:
So, I went and tested this from my phone and computer.  Cleave worked on desktop and iOS Safari, but not with Brave on iPhone.

You are only able to enter digits, not any special characters and it formats pretty well in the textbox.

When viewing the generated button, when clicking on it, the format Cleave creates works well on iOS Safari, bringing the right format number up, but Brave required two popup call dialogs, one with a bad format, then one with the right format ((XXX) XXX-XXXX is the right format for iOS).  It was odd behavior.

PR #105, opened 2018-06-12
POWR-135 implement maxlength on groups and service

joshuami:


joshuami:
Depending on the timing of this one... I may wait to merge until after I've added maxlength to the guide summary—which is dependent on POWR-182.

PR #107, opened 2018-06-13
Hotfix-8 updating composer.lock to latest composer.json changes.

joshuami:


joshuami:
Realized that the composer.lock file was out of sync with our composer.json at the tip of head.

PR #108, opened 2018-06-13
POWR-202 add simpleSAML

jookyg:
I've followed the installation instructions from this page: https://pantheon.io/docs/shibboleth-sso/
I can't get it working locally unless I spend more time trying to configure it for a Lando environment and understanding it better. I'm unable to even get to the welcome page. So I think the best approach now is to get it into a multidev environment using the Pantheon instructions and troubleshoot it there if I still can't get to the welcome page. Just need an approval to get this into the multidev, and I can do another PR after I make some progress getting it running.


web/sites/default/settings.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+$settings['hash_salt'] = "testsalt";
```
joshuami: Nope! We can't put the hash into the settings.php. That should be set in your settings.local.php only.

joshuami:
One more nit picky think, @jookyg. When you create a PR, can you name it the same as the issue or similar? I just renamed this one to "POWR-202 add simpleSAML". It makes it easier for me to understand what each pull request will do before I open it.

PR #109, opened 2018-06-14
Powr 202 updates to SimpleSAML

jookyg:
I need to deploy some updates to this branch to see if I can get SimpleSAML working. I moved the files into a private directory in /web, and also moved the symlink into web. Hoping this will at least let me load the SimpleSAML welcome page.


jookyg:
Having problems, going to delete branch and start with clean copy.

PR #119, opened 2018-06-18
POWR-227 Fix a parameter issue in Slack notification

kxwang:
Test the deploy event on multidev.


kxwang:
Fixed an error with function parameter.

PR #120, opened 2018-06-18
POWR-231 Add a simple workflow and content type updates

joshuami:


joshuami:
Yes. The "in review" state is just an alternate "draft" state for now. Groups would have to decide how they wanted to use that state. We could add some actions that triggered notifications or similar in a future release.

RichardDavies:
This is actually POWR-231 (not POWR-161), right?

joshuami:
@RichardDavies: Yes. I'm not sure where 161 came from. Sadly that commit is not going to be tied to it's issue, but at least the PR will be. Thanks for pointing that out.

joshuami:
@kxwang @RichardDavies @jookyg, could you please add a review?

PR #122, opened 2018-06-19
POWR-214

mike-mcdonald:


joshuami:
@mike-mcdonald, is this a stale PR now? I don't see an open POWR-214 ticket anymore and I feel like this was pulled in during Brittany's POWR-245 work. (Incidentally, I think we should avoid conflating branches like that until everyone on the team is extremely good at creating clean PRs with minimal history. POWR-245 was a mess to merge.)

mike-mcdonald:
Yeah this is no longer needed. POWR-214 was a spike, and contained a change to the active theme, hence why it wasn't merged.

PR #132, opened 2018-06-26
POWR 232 Group content menu

kxwang:


web/sites/default/settings.php
```diff
@@ -53,4 +53,6 @@
   if (is_array($settings)) {
     $settings['trusted_host_patterns'] = array('^'. preg_quote($primary_domain) .'$');
   }
-}
\ No newline at end of file
+}
+
+$settings['hash_salt'] = "testtest";
```
joshuami: @kxwang, you accidentally included a change to the settings.php. This value should be set in your settings.local.php to ensure that you don't need to update a file that can accidentally get pushed to production.

web/sites/default/services.yml
```diff
@@ -55,7 +55,7 @@ parameters:
     #
     # Not recommended in production environments
     # @default false
-    debug: false
+    debug: true
```
joshuami: Similar to the settings.php change, you shouldn't need to update services.yml very often. This value should be set in your settings.local.php to use the development.services.yml file for your local environment. This ensures that we don't push a debugging setting all the way to prod.

kxwang:
OK, I've reverted both accidental changes.

joshuami:
I just pushed a rebase, so you can skip that. I need this code for my columns story POWR-90. I have a feeling it will introduce a lot of merge conflicts to resolve.

kxwang:
Thanks, Josh!

PR #134, opened 2018-06-29
POWR-90 initial commit of columns and new paragraphs approach

joshuami:
Setting up a multidev to show some ideas to others. We shouldn't merge this until we agree on the approach.


web/themes/custom/stormy/templates/field/field--paragraph--field-column.html.twig
```diff
@@ -0,0 +1,10 @@
+{%
+  set classes = [
+    'col-lg'
```
mike-mcdonald: Is it really the intention to stack on tablets?  Frontpage cards don't stack on tablets, so I would think that isn't the desire here either.  Might want to change to 'col-md' or just add it as well.

joshuami: When I did some quick testing with the current Bootstrap-based theme, this worked best. However, I agree that we'll likely want to show these as columns on a tablet when the actual theme is in place. We'll need to consider menu placement along with column breakpoints for an optimal tablet reading experience.

That said, our tablet usage is extremely low compared to phone and desktop devices. I would prioritize their optimization over tablet at the moment—even to the point of giving a tablet a more or less "mobile" treatment.

PR #135, opened 2018-07-02
POWR-238b - Attempt #3 to install SimpleSAMLphp and the associated _auth module

jookyg:


jookyg:
This PR is ready for review. Here is the current status:

* SimpleSAMLphp (SSP) is installed via composer and uses a symlink in the web root "simplesaml"
* SSP config files are located in /vendor/simplesaml/simplesaml/config/ in the repo. I had to override gitignore and store it there, as that's the only location where the _auth module will recognize the config file.
** The CircleCI build process leaves these files alone
** Running composer install locally will overwrite them
* The secretsalt and adminpassword config values are stored in Terminus Secrets in portlandor.powr-238b. Code to parse the private/secrets.json file is located in the SSP config file.
** Two secrets currently stored: simplesaml_secretsalt, simplesaml_adminpassword
** Use "terminus secrets:get portlandor.powr-238b simplesaml_adminpassword" to retrieve the admin passsword
* SSP is configured to only display the initial welcome page without logging in

Issues:

* Because we are managing 2 SSP config files within the /vendor directory (overrides .gitignore), these files get wiped out whenever composer install is run, and care must be taken to revert them and not commit the removal. However, the CircleCI build process seems to leave these files alone.
* If the simplesamlphp_auth module is enabled, the site will break when run locally, because the simplesamlphp config is looking for the private/secrets.json file generated by the Terminus Secrets plugin. Workaround code needs to be put in place that won't crash the site if the secrets file is not found. Until this workaround is in place, the module is not enabled in this branch by default through config; it needs to be manually enabled in the environment until this issue is resolved.

joshuami:
I'm not comfortable with the possibility that `composer install` will wipe out the configuration every time. I think the reason for CircleCI not wiping out the files is that a full composer install is only run on a commit to PR build if the composer.lock changes. We cache otherwise.

If you were to delete the environment from Pantheon and run this from scratch, I'm pretty sure it would just wipe the config directory on that kind of deployment. 😞 

Have we tried adding symlinks? https://www.drupal.org/project/simplesamlphp_auth/issues/2940786#comment-12458832 and the README in the code both point to symlinks as a solution: https://cgit.drupalcode.org/simplesamlphp_auth/tree/README.md

I can't remember if this was talked about before. Also, I'm realizing that README files in Drupal projects are much less easy to find that in a typical Github repo... in part because you only get the README after installing the module via Composer and it is not part of our repo as a result—interesting side effect.

jookyg:
I will test it out with a symlink, but I don't know why that wouldn't get blown away too. The config files aren't overwritten--they're deleted. They don't normally exist in that location, so it must mean that the whole directory is getting deleted and re-created.

jookyg:
@joshuami I verified that composer update will remove anything we put in there, including the symlinks, if the version changes. So unless we specify a version in composer, the next time there's a release, and we run composer update, the config files or symlinks will get blown away again. However, they will still be in Git and just need to be reverted.

I'm still working with Pantheon support. The messages are slow to come back from them, but my open question is whether there's some way we can set the environment variable that allows simplesamlphp to look for its config files in an alternate location.

PR #141, opened 2018-07-09
Hotfix-10 address errors in config import and content admin

joshuami:
The last build created some weird errors on Dev and in my local. These small changes are to address those errors.


joshuami:
@kxwang, would you mind giving me a quick second for this hotfix.

PR #149, opened 2018-07-18
POWR-279 OpenID login lockdown POC

mike-mcdonald:
I removed one of the database logging messages, only logging when the PANTHEON_ENVIRONMENT variable is unset.

This should hopefully run the workflow script to copy in the secrets.

### DIFF
--- a/web/modules/custom/portland/src/Routing/RouteSubscriber.php
+++ b/web/modules/custom/portland/src/Routing/RouteSubscriber.php
@@ -54,7 +54,6 @@ class RouteSubscriber extends RouteSubscriberBase {
       // only log in with an OpenID provider
       if ($route = $collection->get('user.login')) {
         $route->setDefault('_form', 'Drupal\openid_connect\Form\LoginForm');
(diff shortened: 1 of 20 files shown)


web/modules/custom/portland/src/Routing/RouteSubscriber.php
```diff
@@ -2,24 +2,73 @@
...
+
+    // custom overrides on test and live environments
+    if(
+      isset($_ENV['PANTHEON_ENVIRONMENT']) &&
+      in_array($_ENV['PANTHEON_ENVIRONMENT'], ['powr-279', 'dev', 'test', 'live'])
+    ) {
+      // only log in with an OpenID provider
+      if ($route = $collection->get('user.login')) {
+        $route->setDefault('_form', 'Drupal\openid_connect\Form\LoginForm');
+        $route->setDefault('_title', 'Log in with OpenID');
```
joshuami: Let's change this to simply "Log in"

PR #155, opened 2018-07-25
POWR-93 Add Media types: document, image, video.

kxwang:


kxwang:
Ready for review. Lots of changes are repetitive for the three media types. The tricky part is Portland.module which has some custom code to create group media tokens and fix the reset button in media browser.

web/sites/default/config/views.settings.yml
```diff
@@ -9,7 +9,7 @@ ui:
     performance_statistics: false
     preview_information: true
     sql_query:
-      enabled: false
+      enabled: true
```
joshuami: I'm not sure we should be enabling sql_query all the way up to production, but we can pick this up in a future refactor.

kxwang: Sorry, this is a setting slipped in. I turned it on to see the query.

joshuami: I think I merged this before your additional commit. Let me see if I can still merge this change or if I need to have you submit another PR.

joshuami: Hmm. I'll need a separate PR for that last commit. My merge preceded it.

PR #164, opened 2018-07-31
Hotfix-0731 Update config that can't be imported in Live.

kxwang:


kxwang:
@mike-mcdonald Made a commit trying to fix a problem with "changed" timestamp but it didn't work. I've backed out the commit but it already made your previous approval stale. Please review again when you get a chance. Thanks!

kxwang:
The new change is to add a missing DIV tag.

PR #166, opened 2018-08-06
alert-styles: re-implementing alert style from POWR-259 correctly in …

jookyg:
…sass and rebuilding; had previously been written directly to the css.

Also, had to rename this branch to "alert-sty" because the name was too long for multidev. Pantheon only allows 11-char names.


joshuami:
Approved and merging...

PR #174, opened 2018-08-13
POWR-333 add icons to paragraph types.

joshuami:


joshuami:
@jookyg and @brittanysmart, could you give me a quick review on this once it passes the build?

PR #175, opened 2018-08-13
POWR-288 - fixed aggregation bug

jookyg:
Content type field aggregation was grouping on wrong column.


jookyg:
This is fixed and ready for review. It ended up being 2 small changes to the buggy view: added a published filter, and changed the content type grouping column from entity id to target_id.

I've moved the item to QA and assigned to Tracey, with testing instructions in the story. @joshuami it's ready for you to review the PR as soon as it passes QA.

PR #176, opened 2018-08-14
POWR-335 - fixed bug that was preventing notifications from appearing…

jookyg:
… in Menu on services and guides, had to add the Notification plugin to the relationship options in view config.


jookyg:
I have de-selected all content types in the relationship config (this effectively enables all content types), to match the view configuration that is used on the Group landing page. The types that need to be tested are: Event, News and Information. The other 3 types were tested as part of a different bug.

This is ready for review and merge as soon as Tracey completes QA

PR #177, opened 2018-08-15
POWR-291 refactor location content and paragraph types

joshuami:
This PR adds fields to the bureau/office, program, and elected official group types that include contact info. Additionally, it modifies the location paragraph type so that you can optionally add a subparagraph that is either a service location, bureau/office/program, building, or freeform address.


joshuami:
@kxwang and @jookyg, because of the existing location field, this story is kinda hard to get a clean one-time `drush cim` to execute. It needs to be run twice, which is a problem for the CircleCI build and subsequent tests. If you can give me a review, I feel like we are safe to build into master/dev.

PR #179, opened 2018-08-16
POWR-95 Pantheon has SOLR, so let's give that a try

joshuami:


joshuami:
Merged my other test branch with the same stuff.

PR #187, opened 2018-08-22
POWR-361 Enable account creation from OpenID

mike-mcdonald:


mike-mcdonald:
Tested adding my Office 365 account successfully.

PR #188, opened 2018-08-22
POWR-263 Add an icon for external links.

kxwang:


web/sites/default/config/extlink.settings.yml
```diff
@@ -0,0 +1,18 @@
+extlink_target: true
```
RichardDavies: I do not think we should force external links to open in a new window because it's not user friendly. I've provided more info about this in a comment on the Jira story.

kxwang:
FA 5.0 changed file icon names. I've updated the icon name by removing "-o". GitHub doesn't allow patch so I'm using Slack to sent it. Easier to have one person maintain this PR.

See the missing icon here:
https://powr-263-portlandor.pantheonsite.io/bds/documents/checklist-better-writing

joshuami:
Shoot. I need to remove that patch file. No sense muddying up the repo.

PR #191, opened 2018-08-24
POWR-326 Enable media for Project group type.

kxwang:


kxwang:
It appears Media entities are not installed for Project. I installed them and Media showed up for the project POWR.

PR #192, opened 2018-08-24
Hotfix-17 Fixed error when creating group media entities.

kxwang:


kxwang:
The function gmedia_save_group_content doesn't exist. Removing it to avoid getting the error when creating a new document. Updated composer.json to use the new patch file.

Please note Pantheon env seems to be unstable this morning. I had to refresh the page when using the hotfix-17 multi dev site.

PR #194, opened 2018-08-29
POWR-301 Added cronpub scheduling module

RichardDavies:


joshuami:
@RichardDavies, I am closing this PR since we are not taking this approach.

PR #201, opened 2018-08-31
POWR-122 - Programmatic redirects from content edit form

jookyg:
This story adds a custom field to all content and group types (except Alert and Notification) that allows editors to enter one or many paths that should redirect to the current node. Once past validation and the node is saved, these paths are saved as redirects using the Redirect module API.

The functionality has been created in a new module portland_legacy_redirects.


jookyg:
This has been rebased and passed QA

PR #203, opened 2018-09-04
POWR-258 Install and configure linkit module

RichardDavies:


web/sites/default/config/comment.type.comment.yml
```diff
@@ -0,0 +1,10 @@
+uuid: b3553026-e213-470d-b712-55eaac720de7
+langcode: en
+status: true
+dependencies: {  }
+_core:
+  default_config_hash: bqZsN31T2n0UjcbyCpOPi9D2iO0sAOHR7FnEs9qMvaA
+id: comment
+label: 'Default comments'
```
joshuami: I want to remove this comment dependency before I merge this. I'll tackle that tomorrow.

PR #213, opened 2018-09-07
Powr 94 - Audio media type and paragraph type

jookyg:
Should function identically to the Video media type and paragraph type, but with slightly different verbiage.

I have had a curious issue when deploying/building to multidev. I get a build error the first time through. 
I've deleted the pull request and environment and retried to confirm it's repeatable.

```
The import failed due for the following reasons:                         [error]
Unexpected error during import with operation create for group.content_type.project-group_media-audio: The &quot;group_media:audio&quot; plugin does not exist.
Unexpected error during import with operation create for group.content_type.bureau_office-group_media-audio: The &quot;group_media:audio&quot; plugin does not exist.
```
When I re-run the build in CircleCI from the point it broke, it succeeds. This seems to indicate that there is a dependency on one of the config files that hasn't been imported yet, or that they're getting imported out of order. When I run it the 2nd time, the dependent config seems to be there. I don't know what to do about this, but simply re-running the build fixes the problem.


jookyg:
This has passed QA and working as expected.

I'm still experiencing the build/cim issue, but it completes successfully after re-running the CircleCI build, or when importing locally, when re-running cim again.

PR #221, opened 2018-09-12
POWR-402 updated filter field label in entity browser view from "vide…

jookyg:
…o" to "audio."


jookyg:
This is ready for QA and review.

PR #227, opened 2018-09-14
POWR-360 - Update group bulk alias only if shortname changes

jookyg:
I also noticed that group media wasn't included in the bulk alias update, so I added that too.


jookyg:
This is ready for review and merge as soon as it passes QA.

PR #250, opened 2018-09-28
POWR-322 Use taxonomies for alerts and modify templates

mike-mcdonald:


joshuami:
Excellent POC. Thanks, Mike!

PR #262, opened 2018-10-05
POWR-379 - Added default metadata configuration and changed site name to Portland.gov

jookyg:
As discussed in sprint planning, I've taken a quick look at what other cities are using as their site name and how it's formatted. They're all over the place, but the format we discussed seems to be the most common for sites with short, concise domain names ("Page Title | Portland.gov"). I think this is the shortest, cleanest approach and is used by boston.gov, seattle.gov, raleighnc.gov, riversideca.gov, AustinTexas.gov, et al.

This is ready for review & merge as soon as it passes QA.


jookyg:
This PR is ready for review and merge once it passes QA.

jookyg:
Having problems with the multidev site. I'm deleting it and trying a new PR.

PR #272, opened 2018-10-15
POWR-456 update to 8.6 and lightning 3.2

joshuami:


web/sites/default/config/views.view.media.yml
```diff
@@ -3920,16 +3920,7 @@ display:
...
-      menu:
-        type: tab
-        title: Media
-        description: ''
-        expanded: false
-        parent: ''
-        weight: 0
-        context: '0'
-        menu_name: main
+      path: admin/content/media-table
```
RichardDavies: Not sure if this path was supposed to change or not (and both appear to still work). I also just noticed there's two "media" tabs now on the /admin/content page.

PR #274, opened 2018-10-16
POWR-438 Bulk update projects and programs aliases for existing groups.

kxwang:


kxwang:
POWR-438 branch has been merged before. This is some unfinished work that's put into branch powr-438-1.

PR #275, opened 2018-10-16
Powr 379 - Metatag configuration

jookyg:


jookyg:
I had to re-create the environment due to deployment issues. All working as expected. The review is just of the config files for the Metatag module.

PR #278, opened 2018-10-19
POWR-458 Add more behat tests

mike-mcdonald:


tests/features/add_service.feature
```diff
@@ -0,0 +1,27 @@
...
+    And I should see "Related content"
+    And I should see "Legacy path"
+
+    When I press "Add Step"
+    And I wait for AJAX to finish
+    Then I should see "Step"
+
+    When I fill in "Title" with "Test service"
+    And I select "Online" from "edit-field-service-mode-0-subform-field-service-modes"
+    And I fill in "edit-revision-log-0-value" with:
```
joshuami: Is this using the new revision field that @jookyg added in POWR-444? If not, I think this could be problematic.

PR #291, opened 2018-10-29
POWR-485 - output service steps in a more standard way...

jookyg:
...so that we don't need to manually retrieve paragraph field content.

I rejiggered the field templates a little so that we don't need to manually reach into the paragraph field content and output it. Content is handled using default methods, but with some adaptation of the enclosing markup.


web/themes/custom/cloudy/templates/field/field--paragraph--field-service-modes--service-mode.html.twig
```diff
@@ -0,0 +1,17 @@
+{#
```
joshuami: This doesn't seem to execute any code? Why do we need it?

PR #293, opened 2018-10-30
POWR-485 fix a field requirement for document media type

joshuami:


joshuami:
Fixed this with a hotfix instead.

PR #296, opened 2018-10-31
POWR-503 Overwrite title for group-based views to prevent duplicates

RichardDavies:


web/modules/custom/portland/portland.tokens.inc
```diff
@@ -9,38 +9,38 @@ use Drupal\group\Entity\GroupContent;
  * @ingroup portland
  */
 function portland_token_info() {
-  $tokens = [];
+  $info = [];
```
joshuami: Why the refactor of the variable name? Just curious why we would use "info" over "token" here.

RichardDavies: Because that hook defines meta **information** about the tokens so using $info seemed like a more semantic name to me and makes it slightly easier to grok it's purpose (plus it was the variable name used by one of the examples online I came across when I was trying to understand what these functions do.) 

Somewhat of a trivial change, I know, and not required by the implementation I ended up using, but I figured a slight refactor wouldn't hurt since I was already editing that file at one point in my story when trying to create a custom token.

joshuami: Ah, okay. I can live with that.

joshuami:
@jookyg, can we close this PR? It seems like it is not practical to merge this and you already created powr-503a.

joshuami:
Fixed by a different PR.

PR #301, opened 2018-11-02
POWR-524 - extended Moderated Content view to include data export

jookyg:
The data export is currently configured to include the following fields:
* Title
* URL
* Group
* Content type
* Moderation state
* Publication status
* Editor
* Reviewer
* Log message

The Group column and Group, Editor and Reviewer filters have been added to the main view. I've also made more of the columns sortable, which required refactoring which fields are used to display the Group, Editor and Reviewer values. The previous version of the view was sorting by id numbers, not titles/names.

The download link I had previously added manually to the main view header has been removed and replaced with the built-in CSV button below the pagination controls.

The export display has been property attached to the main view so that it correctly remembers and honors filter settings.

**NOTE: There is an issue that concerns me. When I run the export on my local Lando environment, I get a 503 error caused by the Varnish cache unless the view is filtered to only show a subset of items. Is this a memory issue? I haven't seen it occur in multidev env, but if the amount of content grows significantly, it might pop up. There used to be an export batching feature in the D7 version of Views Data Export, but it hasn't been added yet to the D8 module, though there is active development on a patch.**


jookyg:
Rebased as of today at about 2:30 PM.

PR #316, opened 2018-11-17
POWR-592 Set service step card deck cards to layout max two across

brittanysmart:


web/themes/custom/cloudy/scss/components/_card.scss
```diff
@@ -1,19 +1,18 @@
...
-      }
-    @include media-breakpoint-up(lg) {
-      flex-wrap: wrap;
+    &:first-child {
+      max-width: 100%;
+    }
+
+    @include media-breakpoint-up(sm) {
+      flex-basis: calc(50% - 30px);
+      max-width: calc(50% - 30px);
```
mike-mcdonald: The 30px is a magic number, and can be generated based off of the card-deck-margin variable you use below.

web/themes/custom/cloudy/scss/components/_card.scss
```diff
@@ -1,19 +1,18 @@
...
+      margin-bottom: $card-deck-margin;
```
mike-mcdonald: Consider multiplying this by two. The margin between cards is asynchronous horizontally and vertically.  Alternatively, I like trimming the margins down.  I played with making the card margin .5rem and liked the results.  You'd need to override the variable carefully, after looking at bootstrap's definition for it, as it is related to another card variable.

.circleci/config.yml
```diff
@@ -59,6 +59,19 @@ behat_tests: &behat_tests
         key: terminus_-{{ checksum "composer.lock" }}
         paths:
           - "/tmp/vendor"
+    - restore_cache:
```
joshuami: Heads up that you'll need to rebase or merge master to pick up the changes to master from powr-494-1 before this PR should be approved.

PR #330, opened 2018-12-03
POWR-591 - initial config of entity translations vs paragraphs

jookyg:


joshuami:
@jookyg, can we close this PR?

PR #373, opened 2019-01-25
POWR-709 hotfix to update Drupal Core and Lightning

joshuami:
I had to do a bit of clean up after pulling in this latest set of updates. The additional files changed helped with the following:

* Updated the scheduler to only allow scheduled by the hour changes.
* Turned off moderation dashboard redirect.
* Renamed _media.css to _media.scss and added class to fix the preview thumbnail on image upload.

Every other file or configuration change was generated by the updates.


joshuami:
@mike-mcdonald, I felt the same way, but it makes sense. Lightning does not always upgrade to the latest core immediately. By removing core from the explicit requirements, it means we are using the version of core tested with the latest Lightning. This is better practice.

PR #381, opened 2019-01-31
POWR-739 Add drush entup and updb when deploy to TEST and LIVE. Also adjusted drush command sequence in MultiDEV and DEV.

kxwang:


web/private/scripts/drush_config_import.php
```diff
@@ -1,9 +1,13 @@
 <?php
```
joshuami: Should we rename this file to drush_post_deploy.php?

joshuami: This would require an update to pantheon.yml as well, but the note about what the script does there should be updated anyway.

kxwang: Makes sense. Made the update

PR #390, opened 2019-02-01
POWR-719 Implement new body field with embed buttons

jookyg:
The new Body Content field that supports media embed is now implemented on all of the content types that have a Body paragraphs field, and also on the new Page content type.

The old Body field is visible in form display to administrators, but hidden in the default display.

The new custom module portland_media_embed_helper implements a text filter that inserts a media type class to embedded entity containers, and changes the Back button text in the embed dialog.


web/sites/default/config/core.entity_form_display.node.guide.default.yml
```diff
@@ -4,6 +4,7 @@ status: true
 dependencies:
   config:
     - field.field.node.guide.field_body
+    - field.field.node.guide.field_body_content
```
joshuami: Since we are moving this content type into information. We don't need to add the field to the guide content type. (In fact, it might be confusing for migrators.)

web/sites/default/config/core.entity_form_display.node.information.default.yml
```diff
@@ -4,27 +4,29 @@ status: true
 dependencies:
   config:
     - field.field.node.information.field_body
+    - field.field.node.information.field_body_content
```
joshuami: Since we are moving this content type into information. We don't need to add the field to the information content type. (In fact, it might be confusing for migrators.)

web/sites/default/config/moderation_dashboard.settings.yml
```diff
@@ -1,3 +1,3 @@
-redirect_on_login: false
```
joshuami: This is supposed to be false. I think you may have accidentally overwritten some config from master in one of your commits.

web/sites/default/config/lightning_scheduler.settings.yml
```diff
@@ -1 +1 @@
-time_step: 3600
```
joshuami: Here is the other override that was thrown off. This should be 3600. The good news is I think these were probably from running `drush update:lightning` at some point. Let's chat about this real quick and figure out the source before you fix.

web/sites/default/config/embed.button.node.yml
```diff
@@ -15,3 +15,4 @@ type_settings:
   bundles: {  }
   display_plugins: {  }
 icon_uuid: null
+icon_path: null
```
joshuami: Does this mean we need to fix the references to the embed buttons after deployment to live and then copy that down to local to update the config post. We should add that to the deploy notes.

web/sites/default/config/editor.editor.simplified_editor_with_media_embed.yml
```diff
@@ -0,0 +1,65 @@
...
+            - Blockquote
+            - Format
+            - Styles
+        -
+          name: Lists
+          items:
+            - BulletedList
+            - NumberedList
+        -
+          name: Links
```
joshuami: Is this still going to use the LinkIT profile? We need to make sure and test for that in QA.

web/sites/default/config/core.entity_view_display.taxonomy_term.community_actions.search_result.yml
```diff
@@ -29,3 +29,4 @@ content:
     third_party_settings: {  }
 hidden:
   field_fontawesome_icon: true
+  field_show_on_homepage: true
```
joshuami: Why are the new show on homepage fields showing up in your PR? Those are already in place in master. 🤔

web/sites/default/config/core.entity_view_display.node.information.default.yml
```diff
@@ -27,24 +28,28 @@ third_party_settings:
       settings: {  }
     regions:
       content:
+        - content_moderation_control
         - field_summary
-        - field_body
+        - field_body_content
```
joshuami: Since we are migrating away from the information content type, you can leave the body field alone until we've completed that migration and removed the node.

web/sites/default/config/core.entity_view_display.node.guide.default.yml
```diff
@@ -31,21 +32,19 @@ third_party_settings:
     regions:
       content:
         - field_summary
-        - field_body
+        - field_body_content
```
joshuami: Since we are migrating away from the guide content type, you can leave the body field alone until we've finished migration.

web/sites/default/config/core.entity_view_display.taxonomy_term.community_actions.search_result.yml
jookyg: Revert to master

joshuami: No need to revert. While i don't know what caused it, it is not a breaking change. It just adds the field that should have been hidden to hidden. 🤔

web/sites/default/config/editor.editor.simplified_editor_with_media_embed.yml
jookyg: Ahhhh, I missed the LinkIT config for the new editor. I'll add that.

jookyg:
@joshuami  I've resolved all the reported issues, redeployed, and moved the story back into QA. I added a testing instruction step to verify that the LinkIT dialog is used when inserting links.

web/sites/default/config/core.base_field_override.node.service_location.status.yml
```diff
@@ -0,0 +1,22 @@
+uuid: 69db39b1-25ce-4da0-b0ed-6eed98a2c61d
+langcode: en
+status: true
+dependencies:
+  config:
+    - node.type.service_location
+id: node.service_location.status
```
joshuami: @jookyg, why is this field in here?

jookyg: @joshuami That's the Published checkbox, which I think we're disabling in custom code(?). I'm not sure why that might have been added in this case, but it exists on a few other types: Alert, Information, Notification, Page, and Top Search. I've compared the edit forms between those content types and the others, and can't see a difference in functionality. Do you want me to try removing it and see how it behaves?

PR #393, opened 2019-02-07
POWR-725 News content type updates, group homepage block view

joshuami:


joshuami:
1410x200 is the minimum size we would need to do a banner image like what we have in the mockups for group homepages. It is not a standard we would use on any other fields. 

We may rip out this image field as a part of the work to implement focal point. In theory, the editor should be able to select from the media library instead of uploading unique images for every time they feature something. I need to take a look at that in a follow up story.

I think DS will have enough support to remove any layout builder conflicts when layout builder reaches stable. there are already a couple of good issues going that have patches like https://www.drupal.org/project/ds/issues/2966959.

Even with field layout, I think the DS use case of field overrides is super useful in cases where the developer wants to give some of that control to the site builders.

PR #406, opened 2019-02-14
POWR-660 Security headers

RichardDavies:


web/themes/custom/cloudy/package-lock.json
```diff
@@ -4388,12 +4388,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
joshuami: @RichardDavies, your change should not affect the package-lock.json in the theme. I'm removing this file from the commit.

PR #410, opened 2019-02-14
POWR-727 Basic group menus and views for services and page content types

joshuami:


joshuami:
@jookyg, thanks for the review. The DS attributes are display suite. Projects have a date field that we display and elected officials have an official title. In both cases, I wanted to style those differently.

A program should never have a child bureau_office relationship. Programs are lower in the hierarchy.

Popular is styled in a template: web/themes/custom/cloudy/templates/field/field--field-popular.html.twig

PR #426, opened 2019-02-21
POWR-730-1 document type update

joshuami:


web/sites/default/config/core.extension.yml
```diff
@@ -69,6 +69,7 @@ module:
   inline_entity_form: 0
   inline_form_errors: 0
   key: 0
+  kint: 0
```
mike-mcdonald: Not sure this should be checked in

web/sites/default/config/core.entity_view_display.media.document.teaser.yml
```diff
@@ -39,20 +44,56 @@ content:
...
+    type: entity_reference_label
     weight: 1
+    region: content
     label: hidden
     settings:
-      link_to_entity: false
-    third_party_settings: {  }
+      link: false
+    third_party_settings:
+      ds:
```
mike-mcdonald: Aren't these settings better off in a theme template instead of display suite?

joshuami: Actually, our templates are out of control, so I would rather we had some of the basic layout in ds fields than a 200 line twig template with lots of if statements :)

That said, we'll likely refactor a lot that out into field level templates going forward.

web/sites/default/config/core.extension.yml
joshuami: Dang. I'll fix this an repush. sigh.

PR #445, opened 2019-02-27
POWR-772 - Refactor media embed to use display modes embedded100 and embedded50

jookyg:


web/sites/default/settings.php
```diff
@@ -78,7 +78,7 @@
     $config['config_split.config_split.config_prod']['status'] = TRUE;
     break;
   default:  // Everything else (i.e. various multidev environments)
-    $config['config_split.config_split.config_dev']['status'] = TRUE;
```
joshuami: @jookyg, I don't think you meant to change this did you?

PR #466, opened 2019-03-07
POWR-756 Update README.md with new local search instructions

radmosley:
Removed section about Solr


RichardDavies:
@kxwang 	Is the location of the secret files changing? Do we need to include instructions in the Readme for how to obtain the secrets since they're not stored in Git?

kxwang:
No the keys are still under "files/private". We will put both "solr_core_prod.key" and "solr_core.key" on Live and used Config_split to decide which key to use on the current environment. Local setup will need a copy of "solr_core.key" but doesn't need "solr_core_prod.key".

kxwang:
@radmosley Please update the PR title. This is 756. At first I was confused because I'm working on 757 as we speak. :)

radmosley:
@kxwang thanks title updated.

PR #496, opened 2019-03-21
POWR-271 Add featured media field and banner image

brittanysmart:


web/sites/default/config/core.entity_form_display.media.image.default.yml
```diff
@@ -26,12 +27,17 @@ content:
     type: chosen_select
     region: content
   image:
-    type: image_image
+    type: image_focal_point
     weight: 1
     settings:
+      preview_image_style: banner_max_640w
```
joshuami: @brittanysmart, this should be set to an image style that shows the full image. Otherwise, it crops the image and you can set the focal point correctly.

PR #504, opened 2019-03-27
POWR-810 fix twistieup.svg

ccdatapdx:
In our scss components, _tablesort-indicator.scss references ../../../core/misc/icons/004875/twistie-up.svg... which doesn't exist.

We think you should be able to just use site relative paths such as /core/misc/icons/004875/twistie-up.svg.


RichardDavies:
@ccdatapdx 	Your change should have only touched one file but your commit also included the deletion of a couple of simplesaml symlinks -- probably because you weren't running your shell as an administrator when you did your git clone so it couldn't properly create those files. But no worries, I pushed up another commit to restore them.

PR #509, opened 2019-03-28
POWR-872 New lando tooling

RichardDavies:


web/themes/custom/cloudy/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "dev": "webpack --mode development --watch",
+    "devbuild": "webpack --mode development",
```
mike-mcdonald: A more NodeJs way of writing this would be to call this script "build:dev."  I'd prefer that but it is up to you as it changes the scripts and README.

RichardDavies: Thanks for the feedback. I'm willing to change the name to conform to existing conventions. Also, in your opinion, does it make sense to use a build script that uses the development mode instead of production in our local environments? Or is my new script unnecessary?

mike-mcdonald: Yes it does make sense, since for local development we'd probably want to have un-minified versions of the theme assets, especially the javascript.

As we get more advanced and hopefully introduce CSS purging and/or javascript tree shaking to get the assets down to as small as possible, it'll become even more important to do this this way.

PR #511, opened 2019-03-28
POWR-821 Remove Guide and Information content types

radmosley:
I think I solved the issue of the missing block after deleting the content types.


joshuami:
Closing this out as an unmerged story that was moved to a future sprint.

PR #533, opened 2019-04-16
POWR-909 permit content type

joshuami:
Fairly detailed spec implemented here. Also cleaned up some references to information and guide content types—though more will be picked up in Ronnie's story.


web/sites/default/config/search_api_page.search_api_page.search_portland_gov.yml
```diff
@@ -27,11 +27,12 @@ view_mode_configuration:
   'entity:group_program': search_result
   'entity:group_project': search_result
   'entity:node_city_service': search_result
+  'entity:node_development_permit': search_result
   'entity:node_event': search_result
-  'entity:node_guide': search_result
-  'entity:node_information': search_result
   'entity:node_news': search_result
   'entity:node_page': search_result
+  'entity:node_guide': search_result
```
kxwang: I see Guide and Information content types are removed. Should be removed from the search page too.

joshuami: Not a bad idea if it will remove cleanly. I'll update the PR.

joshuami: Hmm. Second thought. It does not seem that I can remove them from the index in its current state. The search api index configuration is based on an exclusion, so it will be cleaner to remove the content types than to update the search api config. Removing the rest of the dependencies can wait for POWR-966.

PR #535, opened 2019-04-17
POWR-889 Fix PHP exception when visiting a non-existing taxonomy term page

kxwang:


.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ save_cache: &save_cache
     - ./web/profiles/contrib
     - ./web/themes/contrib
     - ./web/libraries
-  key: v1-dependencies-{{ checksum "composer.lock" }}
+  key: v1-dependencies-{{ checksum "composer.lock" }}-20190417
```
joshuami: @kxwang, is this pinning the composer.lock file to a particular date? That doesn't seem like something we would want, but I may not full understand what this key is doing.

PR #551, opened 2019-04-26
POWR-824 City Charter content model

joshuami:


joshuami:
About 20 or so of the changed files are just reordering block display. That said, this is another big PR with lots of config change. Thanks to whomever picks it up.

joshuami:
Closing in favor of POWR-825

PR #554, opened 2019-04-30
POWR-825 City Code content model

joshuami:


joshuami:
> * Why is there a clamav config in here, and addition of clamav to core.extension? Is that related to the new content model?

Hmm. I'm not sure why that would change. It might be a config split error. I'll investigate.

> * Is this content type going to be used in the scope of a group? If so, need to update the const GROUP_CONTENT_TYPE_PLURAL in portland.module to include the new content types.

Charter, code, and policies will be a sitewide content type with a special sitewide role to edit the content. It does not belong to a group, but thank you for double checking :)

PR #558, opened 2019-05-02
POWR-1006 fix sitewide editor permissions

joshuami:


.circleci/config.yml
```diff
@@ -203,12 +203,12 @@ run_tests: &run_tests
           sleep 20 # wait 20 seconds for file sync to finish
 
           if [[ $CIRCLE_BRANCH = "master" ]]; then
-            terminus remote:drush portlandor.dev -- cim -y
             terminus remote:drush portlandor.dev -- cr
+            terminus remote:drush portlandor.dev -- cim -y
             terminus remote:drush portlandor.dev -- -y updb
```
RichardDavies: When I did the lando tooling story, I found these sources which advise running updb before cim:

https://drupal.stackexchange.com/questions/254407/order-of-drush-commands-for-automated-deployment/254411#254411

https://twitter.com/weitzman/status/839125800028864514

kxwang: I didn't know that. Sounds like we'll need to follow that sequence of CR, UPDB, CIM, CR.

joshuami: Updated. Please take a look.

PR #559, opened 2019-05-02
POWR-966 removing dependencies for guides and information content types

radmosley:


radmosley:
Could check to see if these files were properly updated?

joshuami:
Other than the merge conflicts, this actually looks really sound and the test site is not throwing errors.

PR #562, opened 2019-05-06
POWR-961 Tweak search config

kxwang:


web/sites/default/config/search_api.server.searchstax.yml
```diff
@@ -20,10 +20,13 @@ backend_config:
     index_timeout: 5
     optimize_timeout: 10
     commit_within: 1000
-    username: null
-    password: null
+    username: kevin.wang
+    password: Lijcar@97229
```
joshuami: Need to remove and reset this password :)

kxwang: Fix committed.

PR #565, opened 2019-05-08
POWR-841 implement entity embed alignment icons

jookyg:


web/themes/custom/cloudy/templates/form/form-element.html.twig
```diff
@@ -79,15 +79,15 @@
         {{ description.content }}
       </span>
     {% endif %}
-    <label{{label_attributes.addClass(label_classes).addClass(input_classes)}}>
```
joshuami: @jookyg, are you sure we want to remove the label element here. It seems like we want to keep form labels, but I want to make sure I understand the intent.

PR #568, opened 2019-05-08
POWR-821 removing information content type

radmosley:


radmosley:
Wanted to run the test suite after removing the first content type to ensure nothing was broke.

web/sites/default/config/key.key.solr_core.yml
```diff
@@ -9,7 +9,7 @@ key_type: authentication
 key_type_settings: {  }
 key_provider: file
 key_provider_settings:
-  file_location: sites/default/files/private/solr_core.key
```
joshuami: This should not have changed. You may need to cache clear before running a cim. You can revert these changes individually and run cim again.

web/sites/default/config/core.extension.yml
```diff
@@ -16,6 +16,7 @@ module:
   chosen_field: 0
   chosen_lib: 0
   ckeditor: 0
+  clamav: 0
```
joshuami: This should not have changed. Please revert.

web/sites/default/config/clamav.settings.yml
```diff
@@ -0,0 +1,15 @@
+enabled: true
```
joshuami: This file should not have changed. Please revert.

jookyg:
I'm marking this PR closed. It got merged but caused an error in Dev in search. There was one field from the deleted content types that hadn't been removed from the search config. I made a hotfix in branch 821-1 and deployed it, and all is working as expected now.

PR #577, opened 2019-05-13
POWR-1015 Fix npm audit vulnerabilities

mike-mcdonald:
Required adding autoprefixer explicitly


mike-mcdonald:
Yeah if dependabot can handle this then it is a better place for it.

PR #587, opened 2019-05-16
POWR-999 composer.json cleanup

RichardDavies:
Removed:
- drupal/hsts (This functionality is also provided by the seckit module)
- drupal/prepopulate
- drupal/restui

Added/Moved to require-dev:
- roave/security-advisories
- drupal-composer/drupal-security-advisories


RichardDavies:
@joshuami 	@kxwang 	Did I miss any packages that should be moved to require-dev?

kxwang:
Not sure if drupal-composer/drupal-security-advisories is needed by some modules or Composer itself. But otherwise looks fine.

One goal of separating Dev and Non-Dev packages is to only deploy Non-Dev packages to Test and Live (Since Dev is the same as Test, we may have to install Non-Dev packages on Dev too.), and deploy all packages to local and Multi-Dev. Not sure if it's under another story. It'll require some CI script changes. May also need to be done in two stories: the first one disabling modules, the second one removing them from Composer.

RichardDavies:
@kxwang I just added the drupal-security-advisories (at Josh's suggestion). It's only necessary on dev and prevents you from adding a composer package with known security vulnerabilities. It's similar to the roave/security-advisories package, but specific to Drupal packages. Since neither of these are Drupal modules, there's no need to disable them before moving from require to require-dev.

PR #600, opened 2019-05-24
POWR-990 Initial stab at configuring group home pages to have Feature…

jookyg:
…d Content and Featured Groups fields, and updated the Featured display on all affected bundles.


web/themes/custom/cloudy/templates/field/field--field-featured-content.html.twig
```diff
@@ -0,0 +1,87 @@
...
+    label_display == 'inline' ? 'mr-1',
+  ])
+%}
+
+{% spaceless %}
+    {% block field %}
+    <div{{attributes.addClass(field_classes).addClass(classes)}}>
+        {% block label %}
+        {%- if not label_hidden -%}
+            <h2><label{{title_attributes.addClass(title_classes)}}>{{ label|render }}</label></h2>
```
joshuami: You can remove <label> from here. Label is really for form display and not output.

web/themes/custom/cloudy/templates/field/field--field-featured-content.html.twig
```diff
@@ -0,0 +1,87 @@
...
+        {%- endif -%}
+    {% endblock %}
+    <div class="card-group--featured">
+    {% block content %}
+    {%- for item in items -%}
+        <div{{item.attributes.addClass(item_classes)}}>
```
joshuami: Remove this div and it might fix the extra div issue. Recommend that we strip this template down to the barest necessities to get the layout we want for now.

web/sites/default/config/clamav.settings.yml
```diff
@@ -0,0 +1,15 @@
+enabled: true
```
joshuami: This doesn't look destructive, but it is already set in the prod config. I'm not sure why it was included.

PR #608, opened 2019-06-03
POWR-829 remove redirects from redirects table when removed from legacy paths

jookyg:


web/themes/custom/cloudy/package-lock.json
```diff
@@ -4194,7 +4194,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
joshuami: Hmm. I feel like everytime someone does a PR lately, they are changing this file. Even if there are no related changes to the theme. 🤔

PR #616, opened 2019-06-14
POWR 893 make admin content views more compact

jookyg:


web/sites/default/config/views.view.admin_group_content.yml
```diff
@@ -262,9 +262,11 @@ display:
           empty_zero: false
           hide_alter_empty: true
           click_sort_column: value
-          type: string
+          type: boolean
           settings:
-            link_to_entity: true
+            format: custom
+            format_custom_true: '<span class="status-published" title="This page is live"><i class="fa fa-circle" aria-hidden="true"></i></span>'
```
joshuami: We should use the terms "published" and "unpublished" for consistency with other places we use published status.

joshuami: Also, using the word "page" for every type of content is a little problematic. Could we change it to "Has published revision" and "Is unpublished"?

web/themes/custom/cloudy/scss/components/_form.scss
```diff
@@ -95,4 +95,36 @@ span.prev-link {
...
+
+.view-admin-group-content .form-check {
+  padding-left: 0;
+}
+.view-admin-group-content .dropbutton-multiple {
+  margin-right: 2em !important;
+  padding-right: 3em !important;
+}
+.view-admin-group-content table {
+  font-size: 0.8em;
```
joshuami: We should probably use the $font-size-sm variable rather than hardcoding this. We have responsive fonts and this wouldn't get hit by those rules. (I'm less concerned about the padding and margin.)

web/sites/default/config/views.view.admin_group_content.yml
jookyg: @joshuami How about, "Has published revision," and, "No published revision?"

I've updated and redeployed.

joshuami: That works for me.

web/themes/custom/cloudy/scss/components/_form.scss
joshuami: Did you try the variable instead of hard coding an em size? How did it look?

PR #647, opened 2019-07-02
POWR-1178 update group node and my content to properly display reviewer

joshuami:


web/sites/default/config/views.view.media.yml
```diff
@@ -4362,7 +4357,7 @@ display:
           exposed: true
           expose:
             operator_id: uid_op
-            label: Author
+            label: Name
```
kxwang: This line would revert the fix for "Author" filter name.

web/sites/default/config/views.view.media.yml
```diff
@@ -7290,7 +7283,7 @@ display:
...
```
kxwang: Should be Author

web/sites/default/config/views.view.media.yml
```diff
@@ -3479,7 +3479,7 @@ display:
...
```
kxwang: Change to Author

web/sites/default/config/views.view.media.yml
joshuami: I must have missed the merge of your story into this branch. Fixed and rebuilding.

PR #648, opened 2019-07-05
POWR-1166 remove group and add advisory group *group type*

joshuami:


web/sites/default/config/views.view.group_advisory_groups.yml
```diff
@@ -258,7 +262,7 @@ display:
           entity_type: group
           entity_field: label
           plugin_id: standard
-      title: 'Advisory groups'
+      title: Projects
```
kxwang: View title should be "Advisory groups".

joshuami: Nice catch. The fix is on its way.

PR #658, opened 2019-07-15
POWR-1104 Location content model refactoring

joshuami:


web/sites/default/config/core.entity_form_display.node.event.default.yml
```diff
@@ -178,6 +180,21 @@ content:
     third_party_settings: {  }
     type: string_textfield
     region: content
+  field_location:
```
kxwang: What's the difference between field_event_location and field_location?

joshuami: field_event_location is a legacy field based on paragraphs to give users the option to add either a service location, building, group contact info, or custom address as an event location. That field has a lot of complexity for querying location, so as a refactor, I’m introducing “Location” that replaces the need for custom addresses as separate from the entity types. In the end, we will only have the ability to add a named location that includes a get directions link by default for events.

PR #661, opened 2019-07-17
POWR-1189 Add Group Creator role and buttons

kxwang:


web/sites/default/config/group.role.elected_official-531c17e3f.yml
```diff
@@ -0,0 +1,17 @@
+uuid: 52487243-578d-426d-9573-f5dd11a01210
+langcode: en
+status: true
+dependencies:
+  config:
+    - group.type.elected_official
+  enforced:
+    config:
+      - user.role.group_creator
```
joshuami: This should not be changed. We create the elected official, program, and bureau/office groups. Making the group creator required means we have to remove ourselves from the group after creation. The required creator should only be on advisory groups and projects.

joshuami: Whoops. I may have misunderstood this code. It looks like the required creator is on another line.

I'm not sure I understand by the user.role.group_creator is on all group configs instead of just the project and advisory group.

PR #662, opened 2019-07-18
POWR-1185 Make "groups" fields required

radmosley:


composer.json
```diff
@@ -97,6 +99,8 @@
         "drupal/telephone_validation": "^2.1",
         "drupal/textarea_widget_for_text": "^1.1",
         "drupal/time_field": "^1.1",
+        "drupal/toc_filter": "^2.0@beta",
+        "drupal/toc_js": "^1.6",
```
joshuami: @radmosley, this is not a clean PR. It has a lot of dependencies from your other story. This story should only have about 12-15 files changed with reference to a required group field. Please update.

radmosley: Reset to an earlier commit and and resubmitted my commits for a cleaner PR.

web/sites/default/config/core.extension.yml
```diff
@@ -17,6 +17,7 @@ module:
   chosen_field: 0
   chosen_lib: 0
   ckeditor: 0
+  clamav: 0
```
joshuami: This should not have changed. We set ClamAV as part of a config split so that it is not enabled locally. Could you revert this file please?

web/sites/default/config/core.entity_form_display.node.location.default.yml
```diff
@@ -5,6 +5,7 @@ dependencies:
   config:
     - field.field.node.location.field_address_or_entrance
     - field.field.node.location.field_geolocation
+    - field.field.node.location.field_group
```
joshuami: Well shoot. This is not going to work. If location is not first added to the groups via content plugin then this new field will cause a ton of issues. 🤔 

For now, let's remove the requirement of the group field on location. I want to give that some more thought.

joshuami:
@radmosley, why did you close this one? Are you planning to open a different PR? The story is currently in QA so I'm a little confused.

PR #683, opened 2019-08-08
POWR-1183 Use shortname to find group notifications

mike-mcdonald:


joshuami:
@mike-mcdonald, looks like the tests are failing. I'm kicking off a rebuild as I don't see why they should from a change like this.

PR #685, opened 2019-08-09
POWR-1221 Council Meeting example

radmosley:


joshuami:
Closing the pull request, but we should keep around the example until the council documents, meetings, and agenda items are refactored because this example site has a lot of sample documents to work from.

PR #688, opened 2019-08-14
POWR-1181 Group editor role and permissions refactor

ccdatapdx:


ccdatapdx:
Note that I changed Marty Member to the new "Editor" group role.

PR #710, opened 2019-08-19
POWR-1181 refactor group permissions

ccdatapdx:


ccdatapdx:
I changed Marty Member to an "Editor" on the /powr group to pass the CI tests.

web/sites/default/config/group.role.elected_official-member.yml
```diff
@@ -12,41 +12,7 @@ audience: member
 group_type: elected_official
 permissions_ui: true
 permissions:
-  - 'access group_media overview'
-  - 'access group_node overview'
```
kxwang: Bureau member has "access group_node overview" and  "access group_media overview". Are they removed for Elected member on purpose?

web/sites/default/config/group.role.project-member.yml
```diff
@@ -14,48 +14,17 @@ permissions_ui: true
 permissions:
   - 'access group_media overview'
   - 'access group_node overview'
-  - 'create group_media:audio entity'
-  - 'create group_media:document entity'
-  - 'create group_media:image entity'
-  - 'create group_media:video entity'
   - 'create group_node:city_service entity'
```
kxwang: Looks like members for Project can edit, which is different from members in other groups.

web/sites/default/config/group.role.elected_official-member.yml
joshuami: Nice catch. Fixed in the next commit.

web/sites/default/config/group.role.project-member.yml
joshuami: This is weird. City Services is not enabled for the project group type. I can't explain why this permission would be here, but it doesn't do anything. I guess I could manually remove it.

joshuami: Manual removal seems to have no side effects.

PR #730, opened 2019-08-23
Update drupal/console to address issue of install path

joshuami:


joshuami:
If one of you three could do a quick PR review, I think this will fix the issue of console trying to install in our project root on some `lando refresh`.

PR #742, opened 2019-08-28
POWR-1387 Fix list style import

brittanysmart:


joshuami:
Um... changing list styles should not require a 47 file update. What is being done in this PR?

PR #743, opened 2019-08-28
POWR-1303 Implement Automatic Table of Contents

radmosley:


web/sites/default/config/rabbit_hole.behavior_settings.node_type_page.yml
```diff
@@ -0,0 +1,11 @@
+uuid: 2e904430-77e2-4303-a7ff-77f3807df626
+langcode: en
+status: true
+dependencies:
+  config:
+    - node.type.page
+id: node_type_page
+action: display_page
+allow_override: 1
```
joshuami: @radmosley, I'm not sure why this was in the config export. Did you touch rabbit hole settings at all?

web/themes/custom/cloudy/package-lock.json
```diff
@@ -8,7 +8,6 @@
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
-      "dev": true,
```
joshuami: I'm not sure why the package-lock.json was touched with this change. Seems like it should not have changed unless you ran an npm update. This story should not have required that should it?

joshuami: We talked about this in standup. The likely difference was running npm from your machine instead of within lando. You should always use `lando npm` commands rather than using locally installed npm. Let me know if that was not the case as we might need to update our lando node containers.

PR #750, opened 2019-09-03
POWR-1361 add manage members to my groups

ccdatapdx:


web/sites/default/config/views.view.my_groups.yml
```diff
@@ -71,6 +71,7 @@ display:
             group_roles: group_roles
             edit_group: edit_group
             nothing: nothing
+            nothing_1: nothing_1
```
kxwang: Probably a little nitpicking: can we change the machine name of "nothing" and "nothing_1" to "manage_members" and "add_content_media"?

joshuami: Huh... didn't catch that when it exported. Interestingly enough, I changed the administrative names for the two custom text fields I created. It turns out that "nothing" is the default machine name for custom text fields in views. (That is kinda lame.)

The only real fix would be to do an update hook to fix the machine name. Which seems a bit extreme to achieve a better name. Manually adjusting the name just disappears on save. :\

I'm going to turn down this change for now and suggest that it is better to have a bad machine name in the automatically created view than add additional code to fix the automatically created config.

PR #752, opened 2019-09-03
POWR-1364 implemented honeypot

jookyg:


web/sites/default/config/lightning_core.versions.yml
```diff
@@ -6,6 +6,7 @@ admin_toolbar_tools: 0.0.0
 adminimal_admin_toolbar: 0.0.0
 allowed_formats: 1.1.0
 anonymous_login: 1.1.0
+antibot: 1.2.0
```
joshuami: Hmm. Looks like this got left behind somehow. Please remove antibot from the lighting_core.versions.yml file as it might cause a conflict.

jookyg: Ah weird. I'm certain I uninstalled it.

PR #753, opened 2019-09-03
POWR-1263 adding address, phone, email to advisory/project groups

ccdatapdx:


ccdatapdx:
Oops! I guess I didn't realize that I was merging all those other branches/updates...let me know if I need to redo this.

PR #760, opened 2019-09-06
POWR-1157 Add social networks, city section, neighborhood fields to Park Facility content type

kxwang:


web/sites/default/config/field.storage.node.field_city_secsion.yml
```diff
@@ -0,0 +1,24 @@
...
+status: true
+dependencies:
+  module:
+    - field_permissions
+    - node
+    - taxonomy
+third_party_settings:
+  field_permissions:
+    permission_type: public
+id: node.field_city_secsion
```
joshuami: Misspelling: should be "field_city_section".

PR #769, opened 2019-09-10
POWR-1060 Add e-File document links

mike-mcdonald:


web/sites/default/config/core.entity_form_display.media.document.default.yml
```diff
@@ -37,6 +57,14 @@ content:
     third_party_settings: {  }
     type: chosen_select
     region: content
+  field_efiles_link:
+    weight: 8
+    settings:
+      placeholder_url: ''
+      placeholder_title: ''
```
joshuami: @mike-mcdonald, I'm not sure we want to give them the option to override the title. It is possible to turn of the link label per link field. I think that would be appropriate for this type of link. Additionally, we may want to restrict the field to only allow links from the efiles domain—which I believe is also a configuration option.

web/themes/custom/cloudy/templates/field/field--field-efiles-link.html.twig
```diff
@@ -0,0 +1,28 @@
+{% set icon %}
+  {% include '@elements/icons/archives.twig' with {
+    "classes": [
+      "h-5",
```
joshuami: We agreed to hold off on adding Tailwind classes for a bit. There are possible conflicts between the Tailwind library and Drupal that I don't have time to unwind at the moment. Is this to set the height and width on the SVG?

web/themes/custom/cloudy/source/_patterns/01-components/links/_icon-link.twig
```diff
@@ -1,8 +1,18 @@
...
+  'items-center',
+  'mb-4',
+]) %}
+{% set icon_classes = icon_classes|default([
+  'flex',
+  'items-center',
+  'mr-2',
+])%}
 <a href="{{ url }}" {{ wrapper_class_markup ? class }}>
-  <div class="flex items-center mb-4">
```
joshuami: This code is going to conflict with POWR-1311. When a document is embedded, we want as few DIVs around the content as possible so that it can be placed inline within a paragraph or in a list (ordered or unordered). @brittanysmart may have something to add based on her story.

brittanysmart: We could make the wrapper element configurable, but this still won't place the document link inline within paragraph text. See [POWR-1311](https://portland.atlassian.net/browse/POWR-1311) for details.

web/themes/custom/cloudy/templates/field/file-link.html.twig
```diff
@@ -0,0 +1,41 @@
...
+ * Available variables:
+ * - attributes: The HTML attributes for the containing element.
+ * - link: A link to the file.
+ * - icon: The icon image representing the file type.
+ *
+ * @see template_preprocess_file_link()
+ */
+#}
+{% set url = file_url(file.uri.value) %}
+{% set text = file.filename.value %}
```
brittanysmart: When embedding a media entity it needs to display the title, not the filename.

PR #790, opened 2019-09-19
POWR-1470 update focal_point

joshuami:


joshuami:
Note: this PR includes a patch that should have been added via POWR-1348 but was missed.

PR #791, opened 2019-09-19
POWR-1352 Render layers from PortlandMaps

kxwang:


web/sites/default/config/core.entity_view_display.node.park_facility.diff.yml
```diff
@@ -0,0 +1,213 @@
+uuid: b28fc8e0-7008-46bd-9c5f-3f18b4da8c69
+langcode: en
+status: false
```
joshuami: @kxwang, I am a little confused by all display modes that are in this PR. Do you know what triggered them to be created? There are no display modes configured for park facility when I look at the multidev for this one. Should we have default, teaser, featured, related, and search result configured so these other display modes don't appear?

web/sites/default/config/field.field.node.park_facility.panelizer.yml
```diff
@@ -0,0 +1,26 @@
...
+entity_type: node
+bundle: park_facility
+label: Panelizer
+description: ''
+required: false
+translatable: true
+default_value: {  }
+default_value_callback: ''
+settings: {  }
+field_type: panelizer
```
joshuami: Was panelizer a forced requirement here? We are not using it anywhere else. I was hoping we would be able to remove it from our enabled modules at some point.

PR #798, opened 2019-09-25
POWR-1476 Project date and empty date/time fields follow up

brittanysmart:


web/themes/custom/cloudy/templates/field/field--field-published-on.html.twig
```diff
@@ -1,7 +1,12 @@
...
-  'icon': icon
+{% embed "field.html.twig" with {
+  'margin': '2'
 } %}
+  {% block field %}
+    {% set field_classes = field_classes|merge(['align-items-baseline']) %}
+    {{ parent() }}
+  {% endblock %}
+  {% block label %}
+    {% set title_classes = title_classes|merge(['mb-0', 'small']) %}
```
joshuami: I like how the label looks a little smaller than the date for this field. Good call. I wonder if we could repeat that pattern in our defaults 🤔

PR #830, opened 2019-10-07
POWR-1520 uninstall unused modules.

joshuami:
panelizer
panelizer_ipe
panelizer_quickedit
switch_page_theme


web/sites/default/config/core.entity_view_display.node.council_document.default.yml
```diff
@@ -99,4 +92,5 @@ content:
     type: entity_reference_label
     region: content
 hidden:
+  group_content: true
```
kxwang: I noticed the group_content field is added to a few site-wide content types. Is this added by the group patch? It doesn't mean these types are now always associated with a group, correct?

joshuami: None of the content is added to a group, but it does update the config of every node type to hide the group_content view by default. This is an unfortunate side effect of the group content field from that patch. It adds the field programmatically to all node types rather than respecting the settings in group plugins.

I still feel like we may replace this patch with a custom module that uses an actual entity field at some point... I just need to think that out a bit. 

For now this change is harmless as it declares a field we are not using as hidden.

PR #838, opened 2019-10-09
POWR-1438 Address intermittent build error

RichardDavies:
I noticed that the majority of the recent "fake" build failures happened right after the waitForThePageToBeLoaded() function/test which had a timeout of 10s. I suspect our failures are caused by intermittent uncached page loads exceeding 10s. So I've bumped the timeout up to 20s hoping that resolves the issue.


RichardDavies:
We can't really QA this before pushing it all the way up the chain. Then the QA will be "wait and see" if we continue to have fake build failures.

PR #854, opened 2019-10-16
POWR-1357 fix bootstrap import order

joekarasek:
https://portland.atlassian.net/browse/POWR-1357

Description:

Improves import order for bootstrap scss assets.

Details:
- So the original setup was not actually incorrect, as per Bootstrap docs. The BS library uses `!default` on its variable declaration which does not override a variable that was already declared.
- However, I took the opportunity here to break up the import of BS into its component pieces. In this setup, I have placed `bootstrap-overrides` after the import of BS's variables, which doesn't effect the end results but is more intuitive (generally people assume the override comes after the declaration).
- Breaking up the BS comps into its individual components lays the groundwork for future performance optimization. The current style bundle size is 243 kb, I bet we can get that lower by removing the styles for the components we are not using.
- I added a bit more structure to style.scss file a-la comments in the vein of ITCSS. I'm hoping this will make it a bit easier to maintain/understand where parts are coming from/add to. I noticed in `overrides.scss` some TODO's about finding better places to place styles, this should help some with that


joekarasek:
@mike-mcdonald re:single vs double... lets get stylelint up and running to catch these things (and auto-fix the easy stuff). I would be happy to set that up.

PR #858, opened 2019-10-17
POWR-1469 Update webform to 5.4.0

mike-mcdonald:


joshuami:
@mike-mcdonald , you have some merge conflicts in your composer.lock. For some reason all of your composer dependencies are switching to http instead of https endpoints as well. There may be an issue with your Composer set up locally.

PR #860, opened 2019-10-17
POWR-1400 - Use computed field as interface for Redirects

jookyg:
There are 3 main classes of changes included in this PR:
* Custom code in the portland_legacy_redirects module to create new computed field
* Config updates to remove field_legacy_paths from content types, group types, and views
* Config updates to enable the computed field field_redirects in applicable content and group types

Branch has been rebased from master as of the morning of 10/24 and conflicts resolved. The Admin Content view has been updated and re-enabled.


web/sites/default/config/views.view.content.yml
```diff
@@ -1,10 +1,9 @@
 uuid: e755469f-111c-4ad4-819a-13c038622717
 langcode: en
-status: true
+status: false
```
joshuami: @jookyg, I found one issue that was missed in tests. Removing a field can affect views where that field exists. It turned off our admin content view. I wonder if we will be able to show the data from a computed field on an administrative view like this. Could you test and let me know how it goes?

jookyg: I've updated and re-enabled the view and verified no others are impacted. I searched through the config files to verify no lingering references to field_legacy_path exist.

Unfortunately it's not possible to show the output from the computed field in views, since the field is configured to not save to the database. Avoiding duplicate data was one of the reasons we made this field computed, but we could revisit that decision. However, we would then need to re-save every page to save the redirects data into the field table.

I also tried joining to the redirects table from views, but it's not possible by default. The Redirects and Nodes entities don't mix. A cursory search didn't turn up any contrib modules or patches to support this.

PR #866, opened 2019-10-22
POWR-1566 composer update, db update, config export

joshuami:


joshuami:
Won't cleanly apply because of config conflict. Simplifying to separate PRs for each module with a database update.

PR #868, opened 2019-10-22
POWR-1415 Park finder search view

kxwang:


web/sites/default/config/group.role.bureau_office-admin.yml
```diff
@@ -26,6 +26,7 @@ permissions:
   - 'create group_node:news entity'
   - 'create group_node:notification entity'
   - 'create group_node:page entity'
+  - 'create group_node:park_facility entity'
```
joshuami: @kxwang, I don't think we should add parks to these permissions. This will make the park facility appear on every bureau/office content page. We really only want a subset of parks folks with that ability.

PR #873, opened 2019-10-23
POWR-1491 Small update on Service Location display

kxwang:


kxwang:
Had to update the Service Location template after the Building to Location migration.

PR #904, opened 2019-11-05
POWR-1543 Add badge to services in /group/services

julegan:
This adds the services badge to the teaser display for services. It also adds new style to the popular badge style so when the two badges stand next to each other they are distinct from each other and that they also don't look "clickable".


julegan:
Whoops. How do I remove specific commits from this PR? There are a few that carried over unintentionally. Eek.

PR #906, opened 2019-11-05
POWR-1545 fix up TOC config for news.

joshuami:


web/sites/default/config/core.entity_view_mode.taxonomy_term.ief_table.yml
```diff
@@ -0,0 +1,10 @@
+uuid: 0261fe8c-401f-4828-8b23-c9ac4d6347e7
+langcode: en
+status: true
+dependencies:
+  module:
+    - taxonomy
+id: taxonomy_term.ief_table
+label: 'Inline Entity Form Table'
+targetEntityType: taxonomy_term
+cache: true
```
RichardDavies: Why is this file part of the PR?

web/sites/default/config/node.type.news.yml
```diff
@@ -15,12 +15,12 @@ third_party_settings:
     title: 'In this article'
     selectors: h2
     selectors_minimum: '0'
-    container: .node
+    container: .text-formatted
```
RichardDavies: Although using .text-formatted makes the TOC appear on News pages, it does't resolve the CSS specificity issue with all of the TOC selectors. For example, look at the TOC headings on https://powr-1545-portlandor.pantheonsite.io/news/2019/10/11/feature-friday-facets-event-searches-and-more now that I've added a h2 to the feedback block. We should limit the TOC to headings only within the body content as there may be other h2's on a page at some point in the future.

web/sites/default/config/core.entity_view_mode.taxonomy_term.ief_table.yml
joshuami: Cleaning up a config conflict and went ahead and included it. :)

web/sites/default/config/node.type.news.yml
joshuami: It's weird... it doesn't pick up the Categories H2 nor does it pick up the Related H2 🤔

RichardDavies: It will pick up _any_ h2 inside of an element with the class text-formatted.

RichardDavies: ...Maybe that's not an issue. But it seems a little too generic to me.

joshuami: Oh... I get it now. It is because of the feedback area specifically because the block there has a textarea. 

I can switch it to the body content class. I'll update the PR.

PR #913, opened 2019-11-07
POWR-1631 groups field refactoring phase 2

jookyg:
**Features:**
* Removes gcontent_field from composer.json
* Adds field_display_groups to the following content types:
  * Calendar item
  * Developer permit
  * Event
  * News
  * Notification
  * Page
  * Service
* Adds portland_groups submodule
  * Prepopulates field_display_groups with current group
  * Limited field options based on group membership or administrator
* Update pathauto patterns to use new field_display_groups instead of group relationship
* Clean up pathauto patterns

**Potential Issue**
When initially running this deploy to multidev, the config import failed for unknown reason after importing the database from POWR-1630. However, logging in and manually re-running the import from the Configuration Synchronization page worked without issue, and subsequent deploys have been fine.

**IMPORTANT**
This branch is dependent on all dependencies to gcontent_field being removed in POWR-1630, or an error will occur in config import. In the multidev environment, the database from 1630 must be used as a starting point.


web/sites/default/config/core.entity_view_display.node.external_resource.default.yml
```diff
@@ -14,17 +14,10 @@ targetEntityType: node
 bundle: external_resource
 mode: default
 content:
-  field_destination_url:
```
joshuami: Uh oh... This should not have been deleted.

web/sites/default/config/core.entity_view_display.node.service_location.teaser.yml
```diff
@@ -57,7 +57,6 @@ content:
 hidden:
   content_moderation_control: true
   field_body_content: true
-  field_floor: true
```
joshuami: Why was this touched here?

PR #915, opened 2019-11-08
POWR-1587 Change end date field template to default

brittanysmart:


web/sites/default/config/core.entity_view_display.node.event.default.yml
```diff
@@ -186,14 +165,6 @@ content:
             classes: {  }
     type: entity_reference_entity_view
     region: content
-  field_neighborhood:
```
joshuami: @brittanysmart, did you remove neighborhood and topics on purpose?

brittanysmart: No, that's a mistake. There should only be a template change.

joshuami: I'm going to fail it then. This is the only file in the PR. :)

PR #929, opened 2019-11-13
POWR-1659 elected official homepage refactor with display suite

joshuami:


web/themes/custom/cloudy/templates/block/block.html.twig
```diff
@@ -36,13 +36,7 @@
       'block-' ~ plugin_id|clean_class,
     ]
   %}
-
```
brittanysmart: The only issue with removing the bottom margin is that blocks will run up against the element below unless that element has a top margin set. See facets https://powr-1659-portlandor.pantheonsite.io/search/bike

joshuami: Well shoot. I still don't like that we have the bootstrap classes on block level elements like that, but I removed the change for now and we can refactor it in another story.

PR #934, opened 2019-11-14
POWR-1612 update permissions to Full HTML text format

joshuami:


web/sites/default/config/system.cron.yml
```diff
@@ -1,6 +1,6 @@
 threshold:
   requirements_warning: 172800
   requirements_error: 1209600
-logging: 0
+logging: 1
```
RichardDavies: Is this intentional and okay to enable?

web/themes/custom/cloudy/scss/components/_form.scss
```diff
@@ -1,19 +1,348 @@
...
+ // START Cloudy-specific components
+
  input {
   &[type='text'],
   &[type='search'] {
     -webkit-appearance: none;
   }
 }
 
-.form--inline {
```
brittanysmart: Removing this extends causes the input buttons on topic pages to wrap to the next like since the form is no longer display flex. https://powr-1612-portlandor.pantheonsite.io/topics/biking

joshuami: Yep. I'm accepting that regression as it is more Drupal default at the moment. I want to come back around on possibly moving those fields to the toolbar.

web/sites/default/config/system.cron.yml
joshuami: I want to turn on cron logging long enough to get that up in production and see our cron times there for a couple of days. It adds about 10 lines to each cron run, so crazy, but not too crazy. :)

PR #939, opened 2019-11-20
POWR-1510 update group events views to use cards

joshuami:


web/themes/custom/cloudy/templates/field/field-group-html-element--group-date.html.twig
```diff
@@ -16,7 +16,7 @@
         {{ display_date }}
       {% elseif start_date is not empty and end_date is not empty %}
         {{ start_date }}
-        <span class="datetime mx-1" aria-label="{{ 'through'|t }}">&ndash;</span>
+        <span class="datetime mx-1 mb-2" aria-label="{{ 'through'|t }}">&ndash;</span>
```
joshuami: @brittanysmart, here was my quick and dirty fix for this issue. I don't like having the utility classes in here, but it works.

PR #948, opened 2019-11-22
POWR-1598 Refactor page template

brittanysmart:


web/themes/custom/cloudy/templates/layout/page.html.twig
```diff
@@ -34,62 +34,87 @@
...
+              {{ page.menu }}
+            </aside>
+          {% endif %}
+          <div class="page-content {{ page.tools ? 'col' }}">
+            {% block page_content %}
+              {{ page.content }}
+            {% endblock %}
+          </div>
+          {% if page.tools %}
+            <aside class="sidebar-right col-sm-12 col-lg-3">
```
joshuami: This doesn't work. You have to put the rendering code within the region or it will print the region even if that region has no content. Refactor so the div and classes are within the regions.

web/themes/custom/cloudy/templates/layout/page.html.twig
```diff
@@ -34,62 +34,110 @@
...
+    'col-sm-12',
+    'col-lg-3',
+    'order-1'
+  ]
+%}
+
+<main id="main-content" role="main" tabindex="-1" data-swiftype-index="true">
+  <section>
+    {% block main_content %}
+      <div class="main-content {{ not is_front ? 'container' }}">
```
joshuami: I like that you fixed the main-content ID with this change to make the skip to content link work correctly!

That said, do we still need the main-content class if we have the ID? Did you keep it because we have dependent CSS?

web/themes/custom/cloudy/templates/layout/page.html.twig
```diff
@@ -34,62 +34,110 @@
...
+              {{ page.menu }}
+            </aside>
+          {% endif %}
+          <div {{ content_attribute.addClass(content_classes)}}>
+            {% block page_content %}
+              {{ page.content }}
+            {% endblock %}
+          </div>
+          {% if page.tools %}
+            <aside {{ aside_attribute.addClass(aside_classes, 'order-md-2 sidebar-right')}}>
```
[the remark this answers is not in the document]
joshuami: I would use a div for now. We might later decide this is a nav element, but I'm torn about that.

web/themes/custom/cloudy/templates/layout/page.html.twig
```diff
@@ -34,62 +34,110 @@
...
+          <div class="col-sm-6" id="footer_first">
+            {{ page.footer_first }}
+          </div>
+          <div class="col-sm-6" id="footer_second">
+            {{ page.footer_second }}
+          </div>
+        {% endif %}
+      </div>
     </div>
-  </div>
```
joshuami: I don't have this extra div in my local. Wonder where that came from. 🤔

brittanysmart: hmm, not sure.

web/themes/custom/cloudy/templates/layout/page.html.twig
brittanysmart: Nothing is dependent on that class so I'll remove it for now.

PR #951, opened 2019-11-22
POWR-1704 extremely stripped back synonyms file

joshuami:


private/solr-conf/synonyms_und.txt
```diff
@@ -11,35 +11,8 @@
...
-business license,business registration,business permit
-business search=>business lookup
-certificate of compliance,certificate of conformance,certificate of conformity
-classes,lessons,instructions,training
-directory,catalogue,roster
-employee,staff,personnel
 garbage,rubbish,waste,trash,debris,litter
-gym,gymns,gymnasium,gynmnasiums,health club,health clubs,fitness club,fitness clubs,fitness center,fitness centers
-internship,internsips,apprenticeship,apprenticeships,traineeship,traineeships,work experience
-job,jobs,career,careers,employment
```
RichardDavies: I'd recommend keeping this one since jobs are our #1 search.

private/solr-conf/synonyms_und.txt
```diff
@@ -11,35 +11,8 @@
...
-job opening,job openings,job posting,job postings,job listing,job listings
-non emergency,non urgent,elective
-parking,car park
-pay bill,pay invoice,pay account,pay notice,pay charge,pay statement
-pay ticket,pay fine,settle ticket
-permit,permits,license,licenses
```
RichardDavies: I'd recommend keeping this one too as community members probably don't know what's a license or permit.

private/solr-conf/synonyms_und.txt
```diff
@@ -11,35 +11,8 @@
...
-police report,police reports,incident report,incident reports
-property tax,property taxes,real estate tax,real estate taxes
-schedule,agenda,plan,program
-short term rental,short term rentals,temporary housing,transitional housing
-summer camp,summer camps,summer activity,summer activities
-swimming,aquatic
-title 33,zoning code
```
RichardDavies: I'd recommend keeping this one too.

private/solr-conf/synonyms_und.txt
```diff
@@ -65,20 +38,3 @@ phb,housing bureau
...
-bball=>basketball
-buisness=>business
-lisence,licens=>license
-complaince,complance=>compliance
-forst=>forest
-paking,oarking=>parking
-tenis=>tennis
-volly ball,vollyball=>volleyball
-volnteer=>volunteer
-witches castle,which's castle=>witch's castle
```
RichardDavies: I'd recommend keeping this one as we know it's a common misspelling.

private/solr-conf/synonyms_und.txt
joshuami: Yes, but the search logs only ever show folks search for job or jobs, which are stems and don't need a synonym. Career doesn't get enough work to count at the moment.

private/solr-conf/synonyms_und.txt
joshuami: This is true, but "permit" can be used as a verb as well as the noun "permit" that is only a loose synonym for license. In our case, the only thing we have called a "license" in the city is associated with a business license, while permit is associated with a ton of other things. I think we are safer not using it at the moment and boosting the things we find could use the other word by including that word in the summary and/or search keywords.

private/solr-conf/synonyms_und.txt
joshuami: After reading through the spelling versus synonym documentation, I think we should add some misspelling optimization rather than putting misspellings into our synonyms. I could be convinced otherwise, but I also feel like we have a lot to learn about how to get that working correctly.

private/solr-conf/synonyms_und.txt
joshuami: The comment is showing me 4 lines, but I'm going to assume it is the last one that you are referring to with this comment. 

I would completely agree if the title of the page that should be the top result for either of those search terms is literally titled "Title 33 Planning and Zoning". I think it will make it to number one without the boost.

PR #961, opened 2019-11-26
fix: twig namespaces paths

EvanLovely:


joekarasek:
Looks great, thanks @EvanLovely

PR #963, opened 2019-11-27
POWR-1664 - refactor policies migration to use new file field for documents

jookyg:


web/modules/custom/portland/portland.module
```diff
@@ -31,7 +31,9 @@ const GROUP_CONTENT_TYPE_PLURAL = [
     "page" => ["pages"],
     "development_permit" => ["permits"],
     "location" => ["locations"],
-    "calendar_item" => ["calendar/archive"]
+    "calendar_item" => ["calendar/archive"],
+    "city_code" => ["code"],
```
joshuami: @jookyg, I'm surprised to see this change in this branch as well as the code/charter branch. Was one branched from the other? Regardless, we don't need this pluralization in either branch.

web/sites/default/config/migrate_plus.migration_group.city_charter_group_content.yml
```diff
@@ -0,0 +1,10 @@
+uuid: be7b63b4-02a3-472d-b37a-f52cb1469741
+langcode: en
+status: true
+dependencies: {  }
+id: city_charter_group_content
```
joshuami: @jookyg, these references to charter should not be in this PR.

web/sites/default/config/views.view.content.yml
```diff
@@ -1143,15 +1143,24 @@ display:
...
-            'group_node:guide': 'group_node:guide'
-            'group_node:information': 'group_node:information'
-            'group_node:news': 'group_node:news'
-            'group_node:notification': 'group_node:notification'
-            'group_node:page': 'group_node:page'
             'group_node:alert': '0'
             'group_node:building': '0'
+            'group_node:calendar_item': '0'
+            'group_node:city_charter': '0'
+            'group_node:city_code': '0'
```
joshuami: Resave this view.

web/sites/default/config/views.view.group_nodes.yml
```diff
@@ -12,6 +12,8 @@ dependencies:
     - field.storage.node.field_service_location_type
     - field.storage.node.field_start_date
     - node.type.calendar_item
+    - node.type.city_charter
```
joshuami: Resave this view.

web/modules/custom/portland/portland.module
jookyg: Resolved

web/sites/default/config/views.view.content.yml
jookyg: Group has been re-saved and no changes were shown. These items are expected here, since the relationship is set to select all content types, same as in POWR-1544.

jookyg: I resaved the view, but there were no changes. This is expected because we changed the group content relationship to use all content types. It lists them all here even if they group content plugins aren't enabled in any group type. This is same as my comment in 1544.

PR #980, opened 2019-12-05
POWR-1721 refactor topic and action list pages optimized for services and permits

joshuami:


web/sites/default/config/block.block.exposedformevents_indexevents_search_index.yml
```diff
@@ -24,6 +24,6 @@ settings:
 visibility:
   request_path:
     id: request_path
-    pages: "/events\r\n/events/past\r\n/events/search\r\n/events/search/*"
+    pages: "/events\r\n/events/past\r\n/events/search\r\n/events/search/*\r\n/events/*"
```
kxwang: Does /events/* supersede /events/past and /events/search?

joshuami: Ah... yes it does. I added the more global path and should have removed the specific ones. I'll fix that now.

joshuami: Fixed... but now I need your review again. 😆

PR #998, opened 2019-12-12
POWR-1778 turn off media add in entity browsers associated with embed buttons

joshuami:


joshuami:
This is a pre-8.8 PR. Since I was in the config, I went ahead and cleaned up the form displays for images in this PR and added the attribution fields to an accordion. that makes it look more extensive than it is :)

PR #1000, opened 2019-12-12
POWR-1750 - Media losing group content relationship when created inline

jookyg:
* Implements hook_entity_insert to manually add a new media entity to a group if it is created in the context of an inline entity form. Works with all media types.
* Reverts temporary changes to entity browsers as workaround while bug was fixed.

### DIFF
--- a/web/modules/custom/portland/modules/portland_groups/portland_groups.module
+++ b/web/modules/custom/portland/modules/portland_groups/portland_groups.module
@@ -75,15 +75,16 @@ function _portland_groups_assign_media_to_group($entity) {
 
   // get group identified in first delta in field_display_groups
   $group_id = $entity->get('field_display_groups')->getValue()[0]['target_id'];
-  $group = \Drupal\group\Entity\Group::load($group_id);
+  if (!is_null($group_id)) {
+    $group = \Drupal\group\Entity\Group::load($group_id);
 
-  $media_group = \Drupal\group\Entity\GroupContent::create([
-    'gid' => $group_id,
-    'entity_id' => $entity->id(),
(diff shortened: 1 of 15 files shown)


web/modules/custom/portland/modules/portland_groups/portland_groups.module
```diff
@@ -53,6 +53,39 @@ function _portland_groups_inline_entity_populate_field_display_groups(&$form, &$
...
+
+/**
+ * Manually assigns a media entity to a group as group content.
+ * The group is identified by the first delta in the media's field_display_groups.
+ * This should only happen in the context of an inline entity form.
+ */
+function _portland_groups_assign_media_to_group($entity) {
+
+  // only do this for inline entity brower
+  $current_path = \Drupal::service('path.current')->getPath();
```
joshuami: How expensive will it be to call a function like this on every page? 🤔

jookyg: I really have no idea. It seems to be a pretty commonly used function, so I assumed it was negligible. It doesn't get called on every page, just whenever a media item is created. I can look into getting the current path/url using php server var.

jookyg: If you're concerned about the performance of that call, I can replace it with $_SERVER['REQUEST_URI'] without much hassle. Though I would expect that path/url operations are pretty lightweight and don't use any db calls or heavy string manipulation.

joshuami: No need to rewrite. I was just musing.

PR #993, opened 2019-12-10
POWR-1481 Add collapsing javascript and classes to tools region

brittanysmart:

PR #984, opened 2019-12-05
POWR-1558a ESLint and StyleLint

joekarasek:

PR #1007, opened 2019-12-16
[POWR-1782] - PL JS Cleanup - Drupal Behaviors attach in PL

EvanLovely:
## [POWR-1782 - PL JS Cleanup - Drupal Behaviors attach in PL](https://portland.atlassian.net/browse/POWR-1782)

This sets up the foundation well for JavaScript by:

- Breaking out independent functionality into it's own js file that's alongside it's scss file for ease of development on a single feature. The main `cloudy.js` just `import`'s each of those in. I did **not** change the contents of the JS functions for the files since git would have a hard time tracking those changes.
- Sets up WebPack to recognize that if we use `import Drupal from 'Drupal'` that we simply get that value from `window.Drupal` upon script load. This let's us use the modern `import` method for pulling in our dependencies instead of the immediate invoking function expression (aka `(function ($, Drupal) { // stuff })(jQuery, Drupal);` Also, when external dependencies and plugins are needed in future the `import` will allow WebPack to reduce it's file size better. This is set up in the `externals` config item in `webpack.config.js`. Done for jQuery too.
- Instead of importing the entirety of Bootstrap JS in, we have a single webpack entry (`bootstrap.js`) that in turn imports each piece of Bootstrap's JS (alerts, tooltips, etc). It still is including all of Bootstrap but this will be much easier to whittle down over time.
- Pattern Lab is set up to emulate the JS environment in Drupal. This is done by using `<script>` tags to pull in the same JS that Drupal does: same jQuery version and core plugins like "once", along with `drupal.js` and `domready` so we can use `Drupal.attachBehaviors` everywhere for consistency and it will fire the `attach` method inside when the page loads. This will help to keep the developer mental model consistent between the two environments and will reduce surprises.
- Setting the foundation in Pattern Lab for good docs for our patterns by including markdown parsing and syntax highlighting inside any `.js-markdown` element. This is **not** pulled into Drupal, only Pattern Lab.
- The main theme asset files of `main.js` and `style.scss` have been renamed to `cloudy.js` and `cloudy.scss` to align with Drupal standards and to provide better developer hints that these are coming from the `cloudy` theme. 
- Setting up asset compiling (both JS & Scss) to use the modern method of "transpire files based on their destined environment" i.e. tell it what browsers you support and it'll figure out what JS syntax needs changing and what CSS vendor prefixes are needed. This is set in the `.browserlistrc` file and is used by `@babel/preset-env` and `autoprefixer`. PS I set it to basically support IE11 and higher along with targets related to US since I doubt you get many non-US visitors. 
- Switched the production flag over to the standard method of setting the env var `NODE_ENV=production` since many minification plugins use that.

### Testing

- Run `npm install && npm start` in `cloudy` folder to install new dependencies, compile assets and start Pattern Lab.
- Open Pattern Lab and notice the lack of console errors
- Open up other Drupal pages where JS is firing and notice it working still

[POWR-1782]: https://portland.atlassian.net/browse/POWR-1782


web/themes/custom/cloudy/src/js/bootstrap.js
```diff
@@ -0,0 +1,12 @@
...
+import 'bootstrap/js/src/carousel';
+import 'bootstrap/js/src/collapse';
+import 'bootstrap/js/src/dropdown';
+import 'bootstrap/js/src/modal';
+import 'bootstrap/js/src/popover';
+import 'bootstrap/js/src/scrollspy';
+import 'bootstrap/js/src/tab';
+import 'bootstrap/js/src/toast';
+import 'bootstrap/js/src/tooltip';
+import 'bootstrap/js/src/util';
```
joekarasek: Good catch on this!

web/themes/custom/cloudy/cloudy.libraries.yml
```diff
@@ -1,9 +1,14 @@
-global-styling:
+bootstrap:
+  # Note: the Bootstrap CSS is currently contained within the `global` library
```
joekarasek: Great commenting

EvanLovely:
> I know there is some work on main.js on other branches and this PR removes that file.

Agreed; if that's the case then we'll need to re-do that work; that'll be super easy though since each js file is basically just the contents of it's own `Drupal.behaviors.XXXX = {}`. Here's the list of currently open PRs that touch the relevant JS files:

- #993 POWR-1481 Add collapsing javascript and classes to tools region 
    - [Adds a new `Drupal.behaviors.sidebarHandler` object](https://github.com/eGovPDX/portlandor/pull/993/files#diff-d0f112b3c8d9474cf4e7f8843d0fa7c8) that was not there before and doesn't touch anything else in `main.js` - so moving will be easy.
- #984 (linting setup)
	- All changes to JS files were done fix `eslint --fix` so doing again is trivial

So it really looks like we just need to get #993 in :+1:

web/themes/custom/cloudy/cloudy.libraries.yml
```diff
@@ -1,9 +1,14 @@
...
+    dist/bootstrap.bundle.js: {}
+
+global:
+  js:
+    dist/cloudy.bundle.js: {}
   css:
     component:
-      dist/style.bundle.css: {}
+      dist/cloudy.css: {}
       https://use.fontawesome.com/releases/v5.2.0/css/all.css:
```
joshuami: FontAwesome doesn't seem to load in Drupal or Pattern Lab.

joshuami: Picking this up in the powr-1783 branch. Resolving for this branch. I'll merge both fixes to master before pushing to Production to prevent regressions.

PR #1017, opened 2019-12-20
POWR-1795 Search Result Pattern

joekarasek:
This creates a "search result" component/twig pattern and documents it in PL. The pattern is not yet hooked up to Drupal. It also does not include/rely on any Bootstrap/Tailwind CSS (with the exception of the badges, which are marked for refactor once that pattern is build out).


web/themes/custom/cloudy/src/components/search-result/search-result.twig
```diff
@@ -0,0 +1,44 @@
+<div class="search-result">
+  <div class="search-result__body">
+    {% if is_not_published %}
+      {# @todo replace the below badge usage with the cloudy baddge component once available#}
+      <span class="badge badge-danger">{{ 'unpublished'|t|capitalize }}</span>
+    {% endif %}
+    {# @todo replace the below badge usage with the cloudy baddge component once available#}
+    <div class="badge badge-pill badge-light">{{ bundle }}</div>
```
joshuami: Not sure we should call this "bundle". That has a specific Drupal meaning and we actually put several different labels into the badge or overline: bundle, project type, page type, news type, event type. It depends on the content, media, term, or group type what we put into that label.

joekarasek: Oh yes, I agree, was just at a loss for what to call that field. Open to suggestions.

joekarasek: I changed this to "type"

PR #1031, opened 2020-01-06
POWR-1817: point ckeditor assets at correct css dist

joekarasek:
Testing Instructions:
- https://powr-1817-portlandor.pantheonsite.io
- confirm ckeditor works as before

### DIFF
--- a/web/themes/custom/cloudy/cloudy.info.yml
+++ b/web/themes/custom/cloudy/cloudy.info.yml
@@ -13,7 +13,7 @@ stylesheets-remove:
   - '@extlink/extlink.css'
 
 ckeditor_stylesheets:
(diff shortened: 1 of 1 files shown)


web/themes/custom/cloudy/cloudy.info.yml
```diff
@@ -13,7 +13,7 @@ stylesheets-remove:
   - '@extlink/extlink.css'
 
 ckeditor_stylesheets:
-  - dist/style.bundle.css
+  - dist/cloudy.css
   - https://use.fontawesome.com/releases/v5.2.0/css/all.css
```
joshuami: I approved this, then thought... shouldn't we switch this to the locally compiled fontawesome files?

PR #1038, opened 2020-01-10
POWR-1481 Drawer (collapsing sidebar for mobile)

joekarasek:
WIP


web/themes/custom/cloudy/src/layouts/drawer/drawer.scss
```diff
@@ -0,0 +1,112 @@
...
+  position: fixed;
+  top: 0;
+  width: 0;
+  z-index: 10;
+
+  @include transition($tray-transition);
+
+  &--position-left {
+    left: -100vw;
+    order: -1;
```
joshuami: Huh... that's kinda slick. Didn't know you could set order this way. 👍

PR #1041, opened 2020-01-13
POWR-1779 Configured for accordion editor form

radmosley:


web/sites/default/config/webform.webform.website_feedback.yml
```diff
@@ -197,7 +197,7 @@ handlers:
     settings:
       states:
         - completed
-      to_mail: support@portlandoregon.zendesk.com
+      to_mail: support@portlandoregon1576198747.zendesk.com
```
joshuami: This file should not be in the the PR. @jookyg , looks like this reared its head again for someone on a newer branch. We should fix early in Sprint 31.

web/sites/default/config_local/webform.webform.website_feedback.yml
```diff
@@ -1,225 +0,0 @@
-uuid: 3a7da01d-48a2-41b6-835f-18f745481403
```
joshuami: We need this file restored or the feedback form will disappear locally. Please work with @jookyg to resolve.

joshuami: Fixed these by merging in master with Greg's config split fix. The change is now only the 2 correct files. When the builds pass, I will merge.

joshuami:
Closing this story in favor of another story that is completing this work.

PR #1046, opened 2020-01-15
WIP ci: up log level

joekarasek:


web/themes/custom/cloudy/pattern-lab/_patterns/03-pages/groups/elected-official.twig
```diff
@@ -17,7 +17,7 @@
   ]
 } %}
   {% block left %}
-    {% include "@elements/02-text/lead-paragraph.twig" %}
```
joshuami: This doesn't seem to have anything to do with the log level. Did I miss something?

PR #1047, opened 2020-01-16
POWR-1873 - updated config split settings to recognize feedback configs

jookyg:


web/sites/default/config/config_split.config_split.config_dev.yml
```diff
@@ -11,7 +11,8 @@ module:
 theme: {  }
 blacklist:
   - clamav.settings
-graylist: {  }
+graylist:
+  - webform.webform.website_feedback
```
joshuami: Ah... so it is this conflg_split config that was causing the config to not see the manually created splits. Good to know. Could you share some notes on this with the team so that we can increase config split understanding?

PR #1049, opened 2020-01-17
POWR-1820 - Service tabs no longer open to the appropriate tab if bookmarked

jookyg:
This is resolved with a small JS update. When we switched from using numbered tabs to named tabs, the underlying script was not updated.


web/themes/custom/cloudy/package.json
```diff
@@ -40,7 +40,7 @@
     "@fortawesome/fontawesome-free": "^5.12.0",
     "@pattern-lab/cli": "^5.4.0",
     "@pattern-lab/core": "^5.4.0",
-    "@pattern-lab/engine-twig-php": "^5.5.0",
+    "@pattern-lab/engine-twig-php": "joekarasek/patternlab-node#feature/twig-php-renderer-options",
```
joshuami: Why does this call Joe's private fork?

PR #1053, opened 2020-01-21
POWR-1756 URL paths for advisory groups, programs, projects with identical names

kxwang:


web/modules/custom/portland/portland.module
```diff
@@ -13,10 +13,9 @@ use Drupal\Core\Cache\Cache;
 // This file concerns itself with setting up tokens with hooks
 include 'portland.tokens.inc';
 
-
-// The plural is used in the list page URL. E.g. "/bds/services"
-// The group list view's path must match the plural form shown here.
-// If the group list view doesn't exist, visiting the alias will get a 404.
+/**
```
joshuami: I like the move to block comments, but Drupal coding standards still suggest a line length of 80 characters—which makes PR review a lot easier as well. https://www.drupal.org/node/1354

web/modules/custom/portland/portland.module
```diff
@@ -198,11 +213,8 @@ function portland_group_content_type_delete(GroupContentTypeInterface $groupCont
 
 /**
  * Implements hook_ENTITY_TYPE_presave().
+ * POWR-475 Changing short name of a group can cause the group URL to be different from the URLs in the sidebar menu. Root cause is that the group URL is set by PathAuto to be clean and unique. But the URL in sidebar menu is formed by Twig code using the short name as is. The solution is to update short name to ensure it's clean and unique when the group is updated.
```
joshuami: "sidebar menu" should be updated to "group menu". We moved that menu inline under group header information.

web/modules/custom/portland/portland.module
```diff
@@ -213,7 +225,7 @@ function portland_group_presave(GroupInterface $group) {
     }
     // Check if this is a new group or an existing group whose shortname has been modified
     if( $originalShortname !== $newShortname) {
-        $cleanShortname = \Drupal::service("pathauto.alias_cleaner")->cleanString($newShortname);
+        $cleanShortname = $newShortname;
```
joshuami: I wonder if we want to take this chance to rename this field to group_path. I'm a little worried that all of our references to shortname don't really make sense in this new context. 🤔

web/modules/custom/portland/portland.module
```diff
@@ -224,20 +236,51 @@ function portland_group_presave(GroupInterface $group) {
     }
 }
 
+/**
+ * A helper function to clean up group shortname.
```
joshuami: This is a slick function.

web/modules/custom/portland/portland.module
```diff
@@ -839,4 +889,44 @@ function portland_form_alter(&$form, &$form_state, $form_id) {
             }
         }
     }
+    else if(substr_compare($form_id, $group_form_prefix, 0, strlen($group_form_prefix)) === 0) {
```
joshuami: I'm thinking this is a lot of code in the overall portland.module. Should we use this as an opportunity to move some of the group functions into a submodule? 🤔

web/modules/custom/portland/portland.module
kxwang: Trimmed block comments

web/modules/custom/portland/portland.module
kxwang: Updated

web/modules/custom/portland/portland.module
kxwang: Probably will do that as another story. This one is getting complicated now. :)

web/sites/default/config/field.field.group.bureau_office.field_group_path.yml
```diff
@@ -0,0 +1,25 @@
+uuid: b430cde1-a011-4728-a181-7b683dd5e538
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.group.field_group_path
+    - group.type.bureau_office
+  module:
+    - custom_add_another
```
joshuami: I wonder why this was a dependency. Doesn't seem like this field would have an "add another" function. 🤔

kxwang: Not sure how it got in there. This is a singleton field.

PR #1056, opened 2020-01-22
POWR-1891 Update tile pattern

brittanysmart:


web/themes/custom/cloudy/pattern-lab/css/pattern-scaffolding.css
```diff
@@ -96,7 +96,7 @@ details {
 .pl-card-deck {
```
joekarasek: Where again are we using this? I see `cloudy-tile-deck` in our examples. Maybe this is something we could safely remove (the pl prefix makes me think its some legacy pattern lab styles we aren't using)

brittanysmart: Good catch. We removed .pl-card-deck from the pattern lab scss.

joshuami:
> This is a good start. I want to discuss with Julie the difference between the solid blue background tile and the white background tile. I believe they are currently attached to different data sources and imply different content. I would like to revisit that design choice and confirm it, or ditch it.
> 
> I also really don't like how the whole card was made clickable. This needs a different technical approach. This is not standard and breaks native browser functionalities (for example, you cannot right click and choose to open in a new tab).

Hmm. I'm not sure this was addressed. At least there are no comments suggesting it was addressed. I'm holding on merging. We should chat tomorrow during the afternoon retro.

brittanysmart:
We confirmed there will be no white Tile as Tile is meant for featured content and groups only. I verified that I can right-click and open Tile in a new tab on FF, Chrome, and Safari.

PR #1082, opened 2020-02-03
POWR-1838 Refactor OpenGraph meta tags

RichardDavies:
* OpenGraph meta tags for all content types and groups now use the entity's featured image (if defined) for Facebook and Twitter cards
* Created new image styles for Facebook and Twitter cards
* og:title and twitter:title uses `[current-page:title]` instead of `[portland:title]`
 because that token wasn't working (and title doesn't need to include domain name because it's included as part of the card)
* Refactored all meta tag definitions to better utilize inheritance (i.e. set appropriate global definitions and removed redundant overrides that used same value as the inherited value, etc.)


web/sites/default/config/metatag.metatag_defaults.global.yml
```diff
@@ -15,8 +15,9 @@ tags:
   title: '[portland:title]'
   og_description: '[node:field_summary]'
   og_site_name: '[site:name]'
-  og_title: '[portland:title]'
```
joshuami: @RichardDavies, why the change from our custom token?

RichardDavies: The custom token wasn't working. For some reason it didn't seem to be returning any value in this context so the `og:title` element wasn't being rendered on the page. I could be wrong, but I think the custom token just appends " | Portland.gov" to the entity's title, and since the cards already include the domain name as part of the card, including this in the link title on the card is redundant anyway.

PR #1086, opened 2020-02-04
POWR-1767 Content completion view

jookyg:


joshuami:
Closing this out and deleting the multidev. A follow up story was created in sprint 33 to come back around on this approach.

PR #1115, opened 2020-02-14
POWR-1799: Fix form inline layout flow issues around select form element

rockymountainhigh1943:
**This PR fixes an issue where inline floated elements containing form select elements are breaking outside of their parent container.**

See this in action here: http://portlandor.lndo.site/pay

**Background:** This issue is occurring because some of the select options contain very long text, and thus force the natural flow of the floated element to expand to the natural size. The work in this PR sets a max-width 100% constraint to prevent that behavior.

**To Test:**

1. **Visit —>** https://powr-1799-portlandor.pantheonsite.io/pay
2. **Verify —>** The select filter in the right hand sidebar is the correct width so that the right side of the select meets with the vertical line created by the layout's right margin. In short — the select input should not overflow outside of the sidebar.


rockymountainhigh1943:
@joshuami — I am unsure why the behat testing failed on my branch. Would you have any idea?

PR #1116, opened 2020-02-14
POWR-1991 setup style-dictionary w all new colors

EvanLovely:
[POWR-1991](https://portland.atlassian.net/browse/POWR-1991)

- Install deps in theme (aka `npm install` in `cloudy`)
- Build the theme
- Look at the design token source at `cloudy/src/design-tokens/*.js` (currently just `colors.js`)
- View outputted scss colors file at `cloudy/src/design-tokens/dist/_design-tokens.scss`
- View PL color demo at "Elements > Global > Color"


joekarasek:
In the process of bug hunting/refactoring, I also broke out the theme build and PL build as their own tasks in CircleCI (they were previously part of the deploy to pantheon).

.circleci/config.yml
```diff
@@ -93,6 +93,14 @@ build_and_deploy: &build_and_deploy
           composer install --no-suggest --no-interaction
           # TODO: should only deploy non-dev packages
           # --no-dev
+    - run:
```
joshuami: I'm approving this, but there is a caching step that I think we might need to revisit in a future story so that we only rebuild the theme when it has changed. 🤔

joekarasek: That's going to be a bit trickier, as we will have to do something like an md5 hash checksum on every file in the theme directory to determine if a change happened (with the dependencies, it just checks the package-lock file). I'm wondering if its really worth it, as the build of the theme and pl takes 17s, and running a checksum on the theme directory then restoring a cache would also take a few seconds. The gain might be less than 10s.

joshuami:
Hmm. Yeah, it isn't worth it for 17s. I'm wondering if we can remove the current code that does theme cache as part of the build process and save some time. Loading something from cache right before you blow it away with a build is probably less than ideal.

I'll have Richard take a look in his build optimization story. Might be worth a developer meetup to talk through it.


________________________________

PR #1118, opened 2020-02-15
POWR-1854: Implement Alert component from Pattern Lab

rockymountainhigh1943:
**This branch implements the Alert component from Pattern Lab and removes the Alert provided through Bootstrap.**

**I also:**
- Merged the relevant contents from the standalone `_alerts.scss` file into the pattern SCSS file to have a single source of truth for alerts
- Added a missing style declaration for `success` alerts
- Fixed some issues with styles that were preventing the heading and contents from expanding to full width
- Lastly, I removed the bootstraps alert styles and renamed the `.cloudy-alert` to simply, `.alert`

**To test:**
1. **Visit —>** https://powr-1854-portlandor.pantheonsite.io/
2. **Verify —>** Alerts appear/function as they should (should be one showing at the top of the window)


joshuami:
This will get completed in POWR-1954.

PR #1123, opened 2020-02-19
DO NOT MERGE - Breaking Code

joekarasek:
This introduces breaking code to webpack. I want to make sure there are no silent errors happening in the theme build process, as that happened on powr-1991.


joekarasek:
Sweet! I figured it out. When chaining commands with `&&`, in the linux environment if the middle command fails it moves on to the next command, and if that succeeds than the whole process is successful. So `cd /var/html/www && npm failing command && cd /var/foo/bar` was not failing. Broke it apart into separate steps so it properly fails.

PR #1139, opened 2020-02-24
POWR-1954 REFACTOR: Remove custom module alert handler

rockymountainhigh1943:
This PR removes the alert JavaScript and handler maintained within the custom `portland` module. This JS was removed in favor of the alerts.js within the theme.

**To Test:**
1. Navigate to http://powr-1954-portlandor.pantheonsite.io/
1. **Verify —>** The alerts still show up and there are no errors within the browser console pointing to a missing script or library


web/themes/custom/cloudy/templates/content/alert/node--alert--featured.html.twig
```diff
@@ -1,6 +1,4 @@
-{% embed 'node--alert.html.twig' with {
-  'dismissible': TRUE
+{% include 'node--alert.html.twig' with {
+  dismissible: TRUE,
+  showTimestamp: FALSE,
```
joshuami: @rockymountainhigh1943, possible quick fix. The template has showTimestamp set as false, but it is still showing for the featured display mode. I was surprised to see that the teaser and feature were inheriting from node--alert.html.twig instead of the component. I thought all the display modes would pull in the component with specific configuration. That may be a misunderstanding of how it should work on my part.

web/themes/custom/cloudy/src/components/alert/alert.twig
```diff
@@ -4,9 +4,9 @@
 {% set changed = changed|default(random(1262332800, 1577865600)) %} {# random timestamp between 2010 and 2020 #}
```
joshuami: Wouldn't this value get set in the demos under pattern lab rather than with a default that if missed would case fake data to be displayed?

web/themes/custom/cloudy/src/components/alert/alert.twig
```diff
@@ -26,7 +26,7 @@
         } only %}
       {% endif %}
     </h5>
-    <div class="cloudy-alert__body">
+    <div class="alert__body">
       {{ content }}
     </div>
     {% if showTimestamp %}
```
joshuami: Interesting... I don't think this particularly timestamp meets our requirements for accessibility. I'm not sure how that got past the earlier review when this was being roles out. could you update the month to be the full month instead of the abbreviation?

PR #1143, opened 2020-02-25
POWR-1806 Add accessibility updates and refactor

brittanysmart:


web/themes/custom/cloudy/src/layouts/drawer/drawer.md
```diff
@@ -0,0 +1,20 @@
+# Drawer
+
+## Props
+
+- `content`: content of the drawer
+- `position`: left (default), right - Does the drawer sit on the left or right. On larger screen, is it a left or right column.
+- `open_text`: string - defaults to "Menu"
+- `open_info`: string - defaults to "Expand side filters"
```
joshuami: My only issue with this language is that it is directional. If I'm using a screen reader, "side" doesn't have the correct meaning. Maybe just update to say "Expand filters" or maybe get more descriptive like "Expand filters have less results on this page".

What do you think?

web/themes/custom/cloudy/templates/layout/region--menu.html.twig
```diff
@@ -1,7 +1,9 @@
 {% if content %}
   {% include "@layouts/drawer/drawer.twig" with {
-    "content": content,
-    "position": "left",
-    "trigger_text": "More"
+    content: content,
+    position: 'left',
+    open_text: 'More',
+    open_info: 'Expand side nav',
```
joshuami: This would have the same feedback as the demo text.

PR #1177, opened 2020-03-10
POWR-2021 Use a forked version of PL uikit

joekarasek:


.lando.yml
```diff
@@ -56,11 +60,6 @@ services:
         - seccomp:unconfined
       command: ["/usr/local/bin/supervisord", "-c", "/etc/supervisord.conf"]
 
-events:
-  post-start:
-    - node: /bin/sh -c "chmod -R 775 /var/www"
```
joshuami: We need this for the Windows machines. Otherwise, their environments won't correctly build the theme.

.lando.yml
```diff
@@ -56,11 +60,6 @@ services:
...
-    - node: echo "Building theme asset files..."
-    - node: cd /app/web/themes/custom/cloudy && yarn run build
```
joshuami: I'm okay with removing the default build, but we need to update our README to explain to the developers that they don't actually have an accurate theme until they've run lando refresh—which shouldn't be a huge burden as they need to run lando refresh to get the latest changes from master for the new branch they are working on. I nearly always to a lando latest and lando refresh after a rebuild anyway.

.lando.yml
joekarasek: I put this back, including the theme build. Both make sense to be there, shouldn't have been removed.

.lando.yml
joekarasek: On the first boot up, I don't think they would have any theme. I have put back this whole block.

PR #1178, opened 2020-03-11
POWR-2094 update EVA Park Map to use field_location

joshuami:


web/sites/default/config/field.field.node.location.field_geolocation.yml
```diff
@@ -15,14 +15,15 @@ third_party_settings:
     weight: 0
     geocode_field: field_address_or_entrance
     reverse_geocode_field: ''
-    hidden: false
+    skip_not_empty_value: 0
     disabled: false
+    hidden: false
     plugins:
-      - portlandmaps
+      - arcgisonline
```
kxwang: We were using PortlandMaps to geocode address which CGID claims has better result.

web/sites/default/config/search_api.index.full_index.yml
```diff
@@ -237,6 +250,14 @@ field_settings:
     dependencies:
       config:
         - field.storage.node.field_history
+  field_location:
```
kxwang: This field is under "Group" not "Content". In `/admin/config/search/search-api/index/full_index/fields`, I tried to added `Location » Content » Geolocation` for Content but can't expand beyond the `field_location -> Content` to reach the Geolocation field.

web/sites/default/config/field.field.node.location.field_geolocation.yml
```diff
@@ -15,14 +15,15 @@ third_party_settings:
...
     dumper: wkt
     delta_handling: default
     failure:
-      handling: empty
```
joshuami: @kxwang, should this be empty or preserve? It seems like preserve would be better.

PR #1194, opened 2020-03-17
POWR-2145: use lock file for node dependencies on circleCi

joekarasek:
Title says it all.

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -116,7 +116,7 @@ tooling:
       - appserver: echo "Clearing cache..."
       - appserver: drush cr -y
       - node: echo "Building theme asset files..."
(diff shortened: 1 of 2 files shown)


.lando.yml
```diff
@@ -116,7 +116,7 @@ tooling:
       - appserver: echo "Clearing cache..."
       - appserver: drush cr -y
       - node: echo "Building theme asset files..."
-      - node: cd /app/web/themes/custom/cloudy && yarn run build
+      - node: cd /app/web/themes/custom/cloudy && yarn run build --frozen-lockfile
```
joekarasek: The build command doesn't do any installs, so this flag doesn't do anything here. I will remove it, but we do want to add it to the start command. I'm going to replace the start command in the other PR as well.

PR #1200, opened 2020-03-19
POWR-2120 path alias patterns for locations and updates to location management

joshuami:


web/sites/default/config/core.entity_view_display.node.service_location.default.yml
```diff
@@ -40,14 +45,33 @@ mode: default
...
+    label: hidden
+    settings:
+      format: default
+      format_custom_false: ''
+      format_custom_true: ''
+    third_party_settings: {  }
+    type: boolean
+    region: content
+  field_floor:
+    type: instagram_embed
```
kxwang: Is the floor field supposed to be "String"?

web/sites/default/config/core.entity_view_display.node.service_location.default.yml
```diff
@@ -40,14 +45,33 @@ mode: default
 content:
   field_body_content:
     type: text_default
-    weight: 4
+    weight: 6
     region: content
     label: hidden
     settings: {  }
     third_party_settings: {  }
+  field_emergency_closure:
```
kxwang: Is there other types of closures? Or will that info be in open hours field?

web/sites/default/config/core.entity_view_display.node.service_location.default.yml
joshuami: Good catch! I'll get that fixed.

web/sites/default/config/core.entity_view_display.node.service_location.default.yml
joshuami: We may add additional closure types later. When everything is normal again, I could see us creating a feature that allowed location managers to set the schedule of open hours including holidays to show when that location is listed. For now, other types of closures could be described in text in the open hours field.

web/sites/default/config/core.entity_view_display.node.service_location.default.yml
joshuami: Fixed.

PR #1201, opened 2020-03-19
POWR-2046 GLOBAL: New typography mixins and helper classes

brittanysmart:
**This PR adds NEW typography tokens to be used in future components and global implementation.**

**IMPORTANT:** The only impact this work has on Beta is that our h1 (besides the homepage h1) will have a font-weight of 500 vs 300. This small change helps reinforce page hierarchy and readability.


web/themes/custom/cloudy/src/design-tokens/typography.js
```diff
@@ -0,0 +1,107 @@
+module.exports = {
+  "fontFamily": {
+    "display": {
+      "value": "'Open Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif"
+    },
+    "prototype": {
+      "value": "'Rubik', 'Helvetica Neue', Helvetica, Arial, sans-serif"
```
joshuami: @rockymountainhigh1943, please remove Rubik from this file. While I know that there is a desire to prototype this font choice, we are not going to be able to introduce font changes until after the COVID-19 response is no longer taking up communications time. I don't want to see the font accidentally get introduced through a pattern implementation and then have to explain it to folks. Just make the prototype value match display for now.

brittanysmart: @joshuami yes, I understand your concern and that we will only introduce a new typeface when the time is right.

I can comment out references to the prototype font, but we will need these variables set up for testing new typefaces as we move forward with design so I'd rather not remove them completely, having to reset them again later as I think this might not be the best use of our time.

The mixin was created to be highly controllable. You must explicitly add `true` when including the typeface mixin inside a selector in order to set the prototype font, so the likelihood of accidentally introducing the prototype font is highly unlikely. Also, we have QA to catch typeface changes.

I'm going to update this PR for you to review. I'm really hoping we can get this merged as it is blocking various tickets.

joshuami: Remove it. Don't just comment it out. It adds complexity that might lead to accidentally introducing a font. Design to Open Sans during the pandemic so I can review work to be merged more easily with no side effects.

brittanysmart: No problem, I've removed it 🙂

PR #1216, opened 2020-03-24
POWR-2160 contact us form to send feedback to ECC from COVID-19 topic

joshuami:


joshuami:
@jookyg, yes... I thought about that. I put it out that far to give us time to come back around and think about retention a bit more. Created POWR-2162 for us to spike out some ideas around retention and detecting whether an email failed to send. We don't really know if an email fails at the moment since we are not comparing the submissions to the tickets created. Definitely something to give more thought 🤔

PR #1218, opened 2020-03-25
POWR-2165 updating queries to use 'now' rather than 'today'

joshuami:


web/sites/default/config/views.view.group_events.yml
```diff
@@ -4335,27 +4350,39 @@ display:
...
+            value: now
             type: offset
           group: 2
-          exposed: false
+          exposed: true
           expose:
-            operator_id: ''
-            label: ''
+            operator_id: field_end_date_value_op
+            label: 'End date (field_end_date)'
```
joshuami: Whoops... this was not supposed to be exposed. Pushing a commit to fix this.

PR #1239, opened 2020-04-01
POWR-2150 Use custom image for Node container

joshuami:


joshuami:
@kxwang , ironically, I couldn't approve this PR since I kicked it off—even though you did all the work :)

PR #1241, opened 2020-04-01
POWR-2185 Update morsel to match news teaser/featured needs

rockymountainhigh1943:
**This PR accomplishes the following:**

- Update the design system morsel pattern to match the latest comps in Figma
- Update the design system demos and documentation to match the new design
- IMPLEMENT: the new morsel pattern within News Teaser & Featured displays (below)
- Exported configuration to the News Teaser & Featured displays (date format updates)

**To Test:**

Visit the below URLs to verify the correct implementation of the new morsel pattern:

- https://powr-2185-portlandor.pantheonsite.io/omf (must be logged in)
- https://powr-2185-portlandor.pantheonsite.io/news
- https://powr-2185-portlandor.pantheonsite.io/novel-coronavirus-covid-19


web/sites/default/config/core.entity_view_display.node.news.featured.yml
```diff
@@ -83,13 +83,13 @@ content:
             fis-def-at: false
             fi-def-at: false
   field_published_on:
-    type: datetime_default
```
joshuami, line 86: Why did you switch this from a date format that we already have? The more correct way to address this would be to update the Month Day Year format.

web/themes/custom/cloudy/src/components/morsel/morsel.twig
```diff
@@ -7,40 +22,20 @@
         classes: ['mb-2'],
       } only %}
     {% endif %}
-    {% if overline %}
-      {% include "@components/overline/overline.twig" with {
-        text: overline,
+    <div class="morsel__heading">
+      {% include "@components/heading/heading.twig" with {
+        url: url,
+        level: 3,
```
joshuami, line 28: Teaser and Search need to be an H2 and Featured needs to be an H3. While these are three different implementations of morsel, the heading level is not the same for every level of morsel. Heading level should not be used exclusively to set size as the headings on the page are how screen readers and keyboard navigators jump from section to section and title to title.

Also, I'm pretty sure we wanted a pretty fixed heading "size" for morsel that was a little smaller than the h3. Ideally, we would pass through the sizing class separate from the heading level at the level of the morsel component.

web/sites/default/config/core.entity_view_display.node.news.featured.yml
rockymountainhigh1943: It's so that the pattern can actually receive a meaningful prop (e.g. string to be formatted by the pattern and inject within our own `<time>` tag inside the pattern) rather than an already rendered chunk of markup (in this case, `<time datetime="March 26, 2020">March 26, 2020</time>` which would then get passed into our `<time>` tag and break everything). When working with design system patterns its pretty standard practice to strip away any tags or rendered markup unless very necessary, then inject your raw data into your pattern where needed.

web/themes/custom/cloudy/src/components/morsel/morsel.twig
rockymountainhigh1943: I'll go ahead and add a `heading_level` prop to morsel so that we can differentiate at the pattern level. Good call! :-)

web/sites/default/config/core.entity_view_display.node.news.featured.yml
joshuami: I'm not sure I agree with this. Instead of having one clear place to go to set time formatting, it now has to be done on every single content config that has date information. That makes it really hard to track down a bug and adds lots of frivolous lines of code. (Not that Drupal is terribly DRY, but this makes it a lot worse.)

I would rather the pattern take the rendered Drupal fields. That way we only override where we need to. Remember, we are not implementing a design system across a lot of sites, we just need a way to theme easier. Let's chat a bit more about this on Zoom tomorrow.

PR #1254, opened 2020-04-07
POWR-2001 Database sanitization and truncation

jookyg:
This update does the following:
* Truncate all webform responses
* Delete all node field revision data from the 9 largest tables, and delete all but the most recent revisions

This results in the database size being cut in half, but it orphans the field revision data from the rest of the node revision tables. The field list is too long and dynamic to write a query to hit them all. If this is unacceptable, an alternative approach would be to do this in a custom module function that can be called by drush when the database is cloned. In a custom module, we could use Drupal functions to get the entity bundles and iterate the fields, and dynamically create the truncation query that way, with no orphans.


web/private/scripts/db_cleanup.php
```diff
@@ -0,0 +1,19 @@
...
+	$watchdog_query = "truncate table watchdog;";
+
+	// Delete webform submission data
+	echo "Deleting webform submissions...\n";
+	passthru('drush sql-query "' . $webform_query . '"');
+	echo "Webform submission deletion complete.\n";
+
+	// Delete log data
+	echo "Deleting log entries...\n";
+	passthru('drush sql-query "' . $watchdog_query . '"');
```
joshuami, line 16: @jookyg, sorry to come up with another issue... but I just thought of another issue with dropping the watchdog logs on all all our multidevs and dev. Just today, Richard needed to look at the watchdog log on dev to see what was happening during a build failure. I don't think we should delete that table. We are only keeping 10,000 log entries now, so it is a lot smaller already.

.lando.yml
```diff
@@ -85,7 +85,10 @@ tooling:
       - appserver: terminus backup:get portlandor.dev --element=database --to=/app/artifacts/database.tar.gz
       - database: echo "Importing database..."
       - database: cd /app && /helpers/sql-import.sh artifacts/database.tar.gz
-      - appserver: echo "Clearing cache..."
+      - appserver: echo "Truncating webforms and logs..."
+      - appserver: drush sql-query "truncate table webform_submission; truncate table webform_submission_data; truncate table webform_submission_log;"
```
joshuami, line 89: We could probably remove this watchdog truncation as well for the same reason as multidev.

PR #1266, opened 2020-04-09
POWR-2203 Send email after locations are updated

kxwang:


joshuami:
Approved after making a small change to the subject line to make it "daily location updates" instead of "daily updates". Seems like that will be easier to identify in all the other notifications we get. Just waiting for the tests to complete to merge.

PR #1269, opened 2020-04-09
POWR-2224 Message styling updates

brittanysmart:
- Update Legacy Path and Back Reference blocks to use default message style
- Update warning message icon
- Clean up message styling
- Add consistent border-radius across messages

### DIFF
--- a/web/modules/custom/portland/modules/portland_edit_forms/back_reference_block.inc
+++ b/web/modules/custom/portland/modules/portland_edit_forms/back_reference_block.inc
@@ -27,11 +27,11 @@ function portland_edit_forms_form_node_form_alter(&$form, $form_state, $form_id)
 
     $form['content_back_reference'] = [
       '#type' => 'container',
-      '#attributes' => ['class' => ['alert', 'alert-info', 'border', 'border-info']],
+      '#attributes' => ['class' => ['message']],
     ];
     $form['content_back_reference']['title'] = [
       '#type' => 'inline_template',
-      '#template' => '<label>{{ title|t }}</label>',
+      '#template' => '<p class="message__heading">{{ title|t }}</p>',
       '#context' => [
         'title' => $view->getTitle(),
       ],
--- a/web/modules/custom/portland/src/Plugin/Block/RevisionBlock.php
+++ b/web/modules/custom/portland/src/Plugin/Block/RevisionBlock.php
@@ -21,8 +21,8 @@ use Drupal\Core\Cache\Cache;
 class RevisionBlock extends BlockBase {
     private static $severity_icon = [
       'danger' => 'ban',
-      'warning' => 'exclamation-circle',
-      'success' => 'info-circle'
+      'warning' => 'exclamation-triangle',
+      'success' => 'check'
     ];
 
     /**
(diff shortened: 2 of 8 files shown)


web/themes/custom/cloudy/src/components/message/message.scss
```diff
@@ -2,42 +2,28 @@
...
 }
 
 .message {
   display: block;
-  margin-bottom: 2px;
   padding: $cloudy-space-4;
-  color: $cloudy-color-neutral-0;
+  color: $cloudy-color-neutral-900;
+  border-radius: $cloudy-space-0;
+  border: none !important; // BAD, but only way to override Bootstrap
```
joshuami, line 18: Do you need the !important if you aren't using a Boostrap class anymore?

web/modules/custom/portland/templates/block--portland-revision-block.html.twig
```diff
@@ -5,10 +5,10 @@
  */
 #}
 {% if current_revision_state %}
-  <div class="portland-revision--block message message-{{ alert_color }} border border-{{ alert_color }} mt-3">
-    {# <h1>{{ title }}</h1> #}
-    <span class="fas fa-lg fa-{{ alert_icon }} fa-fw"></span>
-    <strong>{{ current_revision_state }}</strong> {{ latest_revision_state }}
+  <div class="portland-revision--block message message-{{ alert_color }} mt-3">
+    <span class="fas fa-md fa-{{ alert_icon }} fa-fw"></span>
```
joshuami, line 9: I was surprised to see that this didn't get an SVG for the icons.

joshuami: And I'm also okay with that being a follow up.

web/modules/custom/portland/modules/portland_edit_forms/back_reference_block.inc
```diff
@@ -27,11 +27,11 @@ function portland_edit_forms_form_node_form_alter(&$form, $form_state, $form_id)
 
     $form['content_back_reference'] = [
       '#type' => 'container',
-      '#attributes' => ['class' => ['alert', 'alert-info', 'border', 'border-info']],
+      '#attributes' => ['class' => ['message']],
     ];
     $form['content_back_reference']['title'] = [
       '#type' => 'inline_template',
-      '#template' => '<label>{{ title|t }}</label>',
```
joshuami, line 34: I'm glad you caught this!

web/modules/custom/portland/src/Plugin/Block/RevisionBlock.php
```diff
@@ -21,8 +21,8 @@
 class RevisionBlock extends BlockBase {
     private static $severity_icon = [
       'danger' => 'ban',
-      'warning' => 'exclamation-circle',
-      'success' => 'info-circle'
+      'warning' => 'exclamation-triangle',
```
joshuami, line 24: I'm actually okay with exclamation-circle in this case. The triangle is more of a warning where the circle is more of a notice? I would totally approve this either way, but wanted to get this into place.

brittanysmart: I hear what you're saying. I'm going to create a message component story so we can then swap out message instances for a component. I'll make a note to revisit the icon choice/intention, warning vs. notice usage.

web/themes/custom/cloudy/src/components/message/message.scss
brittanysmart: Nope. It's gone :)

PR #1279, opened 2020-04-15
POWR-2179 Update Code section body text

kxwang:


web/modules/custom/portland_migrations/migrations/city_code_titles.yml
```diff
@@ -13,6 +13,10 @@ process:
     default_value: 1
   title: name
   field_alphanumeric_order: number
+  field_prefix_note/value: note
+  field_prefix_note/format:
+    plugin: default_value
+    default_value: simplified_editor_with_media_embed
```
joshuami, line 19: Oh... that explains all those missing notes. 👍

PR #1284, opened 2020-04-18
POWR-2048 Typography refine vertical rhythm

brittanysmart:


web/themes/custom/cloudy/src/css/_overrides.scss
```diff
@@ -15,11 +15,6 @@ body {
   position: relative;
 }
 
-.lead {
```
joshuami, line 18: @brittanysmart, careful with removing classes from Bootstrap. There were several instances of it that were not in your PR. The use of lead was in config as well as twig templates.

PR #1286, opened 2020-04-20
POWR-1944 POWR-2241 POWR-2253 POWR-2205 combined stories

joshuami:


web/sites/default/config/field.field.group.advisory_group.field_email.yml
```diff
@@ -16,7 +16,7 @@ field_name: field_email
 entity_type: group
 bundle: advisory_group
 label: Email
-description: ''
+description: 'DEPRECATED. This information is being moved to the <strong>contact</strong> field. '
```
kxwang, line 19: Is it safe to remove this field?

joshuami: Not yet. I'm marking the fields as deprecated so that we can help editors migrate to the new contacts field. Once all the data is migrated manually, we can display the new field instead of the old field, then finally remove the deprecated fields.

kxwang: Sounds good.

joshuami:
@kxwang , could you give the another quick look? I had to update an id for the paths and redirects details so that the test could pass. :)

PR #1298, opened 2020-04-23
POWR-2123: Fix News Morsel Pattern

rockymountainhigh1943:
**This PR fixes two issues:**

1) An issue where the Drupal Twig `render` filter was breaking the Pattern Lab Twig compilation process. Here I've simply moved the `render` into the environment where it's actually present (Drupal theme engine).
2) Clears up an annoying console error whenever compiling a watch command (couldn't compile Pattern Lab Morsel demos)

Let me know if you'd like to discuss at all! :-)

**To Test**
1. **Navigate to —>** https://powr-2123-2-portlandor.pantheonsite.io/pattern-lab/?p=viewall-components-morsel
2. **Verify —>** The morsel pattern loads up just fine!
3. **To test the updated functionality —>** You'll need to edit a news item and provide the correct updated date/time, save, then view the list


web/themes/custom/cloudy/src/components/morsel/morsel.twig
```diff
@@ -24,7 +24,7 @@
         classes: ['mb-2'],
       } only %}
     {% endif %}
-    {% if updated_on and (updated_on|render|striptags|date('U') > posted_on|render|striptags|date('U')) %}
+    {% if updated_on and (updated_on|striptags|date('U') > posted_on|striptags|date('U')) %}
```
joshuami, line 27: @rockymountainhigh1943, I noticed that news on https://powr-2123-2-portlandor.pantheonsite.io/news is getting the morsel-updated class even if the updated and posted dates are identical. Is that intentional? Do we need .morsel--updated if the styling associated with an updated date is at the field level?

rockymountainhigh1943:
@joshuami — this should be all set now! The class will only be added when updated_on is correctly set.

PR #1300, opened 2020-04-24
POWR-2235 Multiple reviewers/followers email notifications on all changes to content

kxwang:


web/sites/default/config/rules.reaction.send_email_to_reviewer_after_updating_existing_content.yml
```diff
@@ -39,7 +39,7 @@ expression:
           to:
             - '{{node.field_reviewer.entity.mail.value}}'
           subject: 'Reviewer assignment'
-          message: "<div><a href=\"{{node.revision_uid.entity|entity_url}}\">{{node.revision_uid.entity.name.value}}</a> has assigned you to review <a href=\"{{node | entity_url }}\"> {{node.title.value}} </a></div>\r\n<p/>\r\n<div><h3>Revision log:</h3>\r\n{{node.revision_log.value}}</div>\r\n<p/>\r\n<div>Last updated: {{node.changed.value | format_date('custom', 'l Y/n/j g:i A')}} </div>"
+          message: "<div><a href=\"{{node.revision_uid.entity|entity_url}}\">{{node.revision_uid.entity.name.value}}</a> has assigned you to review <a href=\"/node/{{node.nid}}/revisions/{{node.vid}}/view\"> {{node.title.value}} </a></div>\r\n<p/>\r\n<div><h3>Revision log:</h3>\r\n{{node.revision_log.value}}</div>\r\n<p/>\r\n<div>Last updated: {{node.changed.value | format_date('custom', 'l Y/n/j g:i A')}} </div>"
```
joshuami, line 42: Would it be possible to use /latest for this? I ask because I'm doing a permissions audit right now and considering removing the ability for a non-publisher from seeing any revisions other than the last revision or the published revision. This story will make a difference in whether I lock down that permission or not. 😄

kxwang: I was trying to use Latest first. But I ran into a problem that the "latest" URL is only valid after the node is published once. If the node is never published, I can't get to the latest revision using the Latest URL path.

joshuami: Well shoot. I guess that settles my permissions review then. 😞 

I really wish I had revision level permissions on groups rather than only sitewide. Another option would be to add a sitewide reviewer role that gave access to revisions and determined whether a user could even be added as a reviewer. I don't really want to maintain that, but I'm kinda running out of ideas to align the security of unpublished content and the need to be able to review content. Sigh.

kxwang: Not sure if there is already a module for it. When accessing any revision, we can implement a hook to deny access if the user is not the reviewer. Of course we'll need to make sure it still allow all group members to access revisions.

PR #1301, opened 2020-04-24
POWR-2279 events on content and locations with bonus refactoring

joshuami:


web/themes/custom/cloudy/templates/content/location/node--location.html.twig
```diff
@@ -30,5 +30,9 @@
     {% if view_news > 0 %}
       {{ drupal_block('views_block:news_on_content-block_1') }}
     {% endif %}
+    {% set view_events = drupal_block('views_block:events_on_content-block_2')|length %}
```
brittanysmart, line 33: How can we apply a class to the block heading?

joshuami: There are lots of ways, but most of them involve either a custom template per view type or applying the class to every view block. 

This is another example where I would rather target an h2 or h3 using a class on the view's top level div. That could be applied to each view with 2 clicks rather than adding a ton of additional twig templates.

We've really got to get better naming conventions for our views and using what Drupal gives us rather than overriding every template with custom classes from the pattern lab. It truly makes the Drupal side of things extremely messy.

joshuami: Or even better, we could put a class onto the right side region and set the heading sizes and margins for the entire cascade the same way we would set the cascade for body content. 🤔

PR #1322, opened 2020-05-05
Hotfix-25 Re-enable Rules

kxwang:


web/sites/default/config/lightning_core.versions.yml
```diff
@@ -200,7 +200,7 @@ responsive_image: 0.0.0
 rest: 8.6.2
 rh_group: 1.0.0-beta6
 rh_node: 1.0.0-beta4
-rules: 3.0.0-alpha4
+rules: 0.0.0
```
joshuami, line 203: Huh... this might have been part of the issue. 🤔

PR #1323, opened 2020-05-05
POWR-2237 anonymous user redirects for editors and admin screens

joshuami:


joshuami:
yeah... I realized that the subpathauto changes were actually preventing about half the rules from working. It's a regression that few of us realized because we usually come into the site from the homepage rather than links that other editors send us. 🤷 

The nice thing is this actually simplified our rules a bit. 😄

PR #1335, opened 2020-05-12
POWR-2321 Reviewer email follow ups

RichardDavies:


web/modules/custom/portland/portland.module
```diff
@@ -813,123 +803,115 @@ function portland_cron(){
...
+            "</a> . You will be notified of all changes to this content.<p/>". 
+            "<a href=\"".$node->revision_uid->entity->url()."\">".$node->revision_uid->entity->label().
+            "</a> ".$status_message." a new <strong>".$moderation_state_message[$latest_rev->moderation_state->value].
+            "</strong> revision at ".
+            \Drupal::service('date.formatter')->format($node->changed->value, 'custom', 'l Y/n/j g:i A').
+            ".<p/>".$revision_log_output;
+        
+        if ($reviewers == NULL) return;
+        $current_user_email = \Drupal::currentUser()->getAccount()->getEmail();
+        foreach ($reviewers->referencedEntities() as $reviewer) {
```
joshuami, line 853: Could we add a conditional check here to skip the current revision editor from being notified?

RichardDavies: I suppose so, if you want to expand the scope of the story to include this feature.

joshuami: Nice!

PR #1346, opened 2020-05-18
POWR-2305 - Event cloning

jookyg:


web/sites/default/config/lightning_core.versions.yml
```diff
@@ -238,6 +239,7 @@ transliterate_filenames: 1.2.0
 twig_tweak: 2.0.0
 typed_data: 1.0.0-alpha3
 unified_twig_ext: 0.0.0
+unused_modules: 1.0.0
```
joshuami, line 242: @jookyg I think this might have been an artifact from your testing modules. Would you like to remove it?

PR #1370, opened 2020-05-27
POWR-2343 views to manage content owned by multiple groups

jookyg:
* Added views_field_view module
* Created view to display all content with multiple owners (/content/dupe-group)
* Created subquery view to display all owning groups associated with an individual node


joshuami:
Closing this pull request and closing out the multidev. We can reopen if we think we want to merge this at some point.

PR #1372, opened 2020-05-27
POWR-2235a Try out core_views_facets

kxwang:


joshuami:
Closing this PR out as the search facets seem like the more likely approach. Also... noticed this is the wrong branch name, so editing that for the future. 😄

PR #1379, opened 2020-05-30
Global Header/Navbar/Search/Footer

rockymountainhigh1943:
**The multi-dev environment https://powr-2291-portlandor.pantheonsite.io/ contains all of the recent work for the global header/navbar and global footer. Please use this environment to conduct all testing.**

Instructions for testing, as well as the outlined go-live procedures, are located within the story description of https://portland.atlassian.net/browse/POWR-2289


web/themes/custom/cloudy/src/css/_pages.scss
```diff
@@ -39,4 +39,11 @@ h1.homepage-title {
   h3 {
     @include type-style($cloudy-heading-5);
   }
-}
\ No newline at end of file
+}
+
+// Applies sizing to headings in footer.
+.cloudy-footer {
```
joshuami, line 45: @brittanysmart , this is what I was talking about to reduce the number of block overrides. If we avoid applying the `class="h6"` style utility classes, and use a simple cascade. We can style every header in the footer with 4 lines of code rather than adding a block template or a hardcoded header to the page template.

This creates a smaller PR that is much easier to read and let's developers add blocks without having to do theming.

We need to move towards this approach on blocks and fields to make it easier to use Drupal to do what it does best.

PR #1397, opened 2020-06-08
POWR-2425 Fix park finder autocomplete error

kxwang:


joshuami:
Rolled this change back, but I want to return to it at some point, so I'm going to keep the branch after closing.

PR #1408, opened 2020-06-10
POWR-2413 Re-export code/charter/policy

kxwang:


web/sites/default/config/editor.editor.simplified_editor_with_media_embed.yml
```diff
@@ -20,6 +20,7 @@ settings:
             - RemoveFormat
             - Bold
             - Italic
+            - Underline
```
joshuami, line 23: I'm going to remove the UI for underlining. That is a horrible anti-pattern for web pages that would quickly proliferate if all our editors could use it. I'm okay with underlines that are required by code or charter import to come in, but we need to work with the auditor to remove that from the website since it will trick folks into thinking content is links.

PR #1430, opened 2020-06-17
POWR-2473 Add local Solr service

kxwang:


web/sites/default/settings.php
```diff
@@ -125,6 +125,10 @@
 else if($env == 'live') {
   $config['search_api.server.searchstax']['backend_config']['connector_config']['core'] = 'Production';
 }
+else if($env == 'lando') {
+  $config['search_api.server.searchstax']['backend_config']['connector_config']['host'] = 'solr';
+  $config['search_api.server.searchstax']['backend_config']['connector_config']['core'] = 'index';
```
RichardDavies, line 130: I don't think you need these manual overrides here since you've got a config split for lando local environments. You should be able to just put these values directly in web/sites/default/config_local/search_api.server.searchstax.yml

web/sites/default/config_local/search_api.server.searchstax.yml
```diff
@@ -0,0 +1,66 @@
...
+    - search_api_solr.solr_request_handler.request_handler_extract_default_7_0_0
+    - search_api_solr.solr_request_handler.request_handler_mlt_default_7_0_0
+    - search_api_solr.solr_request_handler.request_handler_select_default_7_0_0
+    - search_api_solr.solr_request_handler.request_handler_spell_default_7_0_0
+    - search_api_solr.solr_request_handler.request_handler_suggest_default_7_0_0
+    - search_api_solr.solr_request_handler.request_handler_tvrh_default_7_0_0
+  module:
+    - search_api_solr
+id: searchstax
+name: SearchStax
```
RichardDavies, line 30: This could be renamed since it's a local server and not Searchstax.

web/sites/default/config_local/search_api.server.searchstax.yml
```diff
@@ -0,0 +1,66 @@
...
+description: 'Solr server hosted by SearchStax.com. The key module sets the Solr host, Solr core, and HTTP auth parameters from secret .key files.'
```
RichardDavies, line 31: Likewise description needs updated.

web/sites/default/config_local/search_api.server.searchstax.yml
```diff
@@ -0,0 +1,66 @@
...
+backend: search_api_solr
+backend_config:
+  connector: standard
+  connector_config:
+    scheme: http
+    host: solr
+    port: 8983
+    path: /
+    core: index
```
RichardDavies, line 40: Set server values here instead of in settings.php.

web/sites/default/settings.php
kxwang: Removed the host name override but keeping the core override. The key file appears to override the Core name.

web/sites/default/config_local/search_api.server.searchstax.yml
kxwang: Updated

web/sites/default/config_local/search_api.server.searchstax.yml
kxwang: Description updated

web/sites/default/config_local/search_api.server.searchstax.yml
kxwang: Done

web/sites/default/settings.php
RichardDavies: Hmm. you're right. But the key file also overrides the host so now I'm getting an error that Local Solr is unreachable and it's trying to reach http://ss626374-s4p42op1-us-west-2-aws.searchstax.com:8983/solr/#/Local. Is there a way to disable the key overrides in the local config split?

kxwang: Ah, that would require we create another key file like "solr_host_local.key" and add a config split for the key setting that points to that file. Similar how we did it for solr_core_prod.key

RichardDavies: Yeah, I suppose that would work. Or you could just revert back to overriding both in settings.php. I'll let you decide which route you think is better.

kxwang: Updated settings.php to set "host" directly.

PR #1445, opened 2020-06-25
POWR-2484 BDS News Pages migration

RichardDavies:


joshuami:
See my notes in POWR-2484. Migrating each PDF into a unique page would kill search. We need to rethinking this one.

PR #1446, opened 2020-06-25
POWR-2493 Set the default group in "field_display_groups" in entity b…

kxwang:
…rowser form


web/modules/custom/portland/portland.module
```diff
@@ -743,8 +742,15 @@ function portland_form_alter(&$form, &$form_state, $form_id)
...
         'node_news_form',
+        'node_news_edit_form',
         'node_page_form',
+        'node_page_edit_form',
         'node_event_form',
+        'node_event_edit_form',
+        'node_construction_project_form',
+        'node_construction_project_edit_form',
+        'node_development_permit_form',
+        'node_development_permit_edit_form',
```
joshuami, line 753: Can we add this to service and service location as well?

kxwang: Added here

PR #1451, opened 2020-06-29
POWR-2488 DB Search based view for News and Services/Pages

kxwang:


web/modules/custom/portland/portland.module
```diff
@@ -799,6 +799,13 @@ function portland_form_alter(&$form, &$form_state, $form_id)
     if (isset($form["replace_file"])) {
         $form["replace_file"]['keep_original_filename']["#access"] = false;
     }
+
+    // For exposed search form, force it to search within the group.
```
joshuami, line 803: @kxwang , I think you can remove this as all the exposed forms are inline instead of placed as a block now.

web/sites/default/config/views.view.group_services.yml
```diff
@@ -423,7 +423,7 @@ display:
         metatag_display_extender:
           metatags:
             robots: noindex
-      path: group/%group/services
+      path: group/%group/services-old
```
joshuami, line 426: Why did you keep this one but delete the news view?

kxwang: View removed

web/modules/custom/portland/portland.module
kxwang: Code block removed

PR #1467, opened 2020-07-08
POWR-2526 Use entity ref view on  charter/code parent field

joshuami:
By using an entity reference view for the autocomplete selection on field_code_parent and field_charter_parent, a non-administrator can still select unpublished content in the entity reference. It is important to turn off SQL rewriting to prevent persisting permission issues.


joshuami:
I thought about adding the filter for charter and code level, but I want to hold off until I've done the zoning code work. We may need to add another layout of "subsection" to support the Zoning Code. I didn't want this to accidentally break if I do that work 😄

PR #1474, opened 2020-07-14
POWR-2388 remove development permit content type PR1

joshuami:
First PR of several to remove unused dependencies and refactor portlandor to have less cruft.


joshuami:
Yep. That patch was fixed in Drupal 8.9.

PR #1485, opened 2020-07-21
POWR-2512 - follow-up updates to Construction/Projects facets

jookyg:
Unfortunately adding the facet blocks changed the weights of most of the blocks in the Tools regions, so there are a ton of files that changed. You'd think it would increment the weight to be the next highest number when adding, but instead it seems to just stick them somewhere in the middle, which affects everything beneath.

We discussed not auto-expanding facet hierarchies, but for the sake of just getting this story done with consistency, I've made the group construction and projects facets expanded. There is another story on the board that deals with auditing all the facets and selecting the appropriate settings. At the moment, there is not enough content for expanded hierarchies to be much of an issue, and it will be solved by the other story before there is.

I also had to add some group fields to the search index config that were needed in order to convert the views to use the database index.

* Converted the views for group construction (content) and group project (sub-groups) to use database index and split each view into 3 subviews: Active, Upcoming, and Completed.
  * DB Search: Group: Construction
  * DB Search: Group: Projects
* Created 2 new views/blocks for the sub-menus in the above views, to mimic the pattern used in DB Search: Group: News.
* Disabled caching for the 
* Added Type, Neighborhood, Topic, and Status facets to each of the 6 new subviews; added Year facets to 2 "completed" subviews.
* Added several group fields to the search API which were in the Solr config but not the db index config.
* Made the construction and project facets auto-expand the topic and neighborhood facets for consistency; the appropriateness of this setting will be evaluated in POWR-2568, but at the moment there is not enough content for it to be a problem.
* Removed the word "Projects" from the sitewide construction and projects view titles.


web/sites/default/config/views.view.group_construction_menu.yml
```diff
@@ -0,0 +1,309 @@
...
+          table: views
+          field: nothing
+          relationship: none
+          group_type: group
+          admin_label: ''
+          label: ''
+          exclude: false
+          alter:
+            alter_text: true
+            text: "<div class=\"sr-only\">{{ label }} News Menu</div>\r\n<ul class=\"nav nav-group-menu\">\r\n<li class=\"nav-item\"><a href=\"/{{ field_group_path|render|striptags|trim }}/construction\" class=\"nav-link{% if path('<current>') matches '[construction$]' %} active{% endif %}\">Active</a></li>\r\n<li class=\"nav-item\"><a href=\"/{{ field_group_path|render|striptags|trim }}/construction/upcoming\" class=\"nav-link{% if path('<current>') matches '[construction/upcoming$]' %} active{% endif %}\">Upcoming</a></li>\r\n<li class=\"nav-item\"><a href=\"/{{ field_group_path|render|striptags|trim }}/construction/completed\" class=\"nav-link{% if path('<current>') matches '[construction/completed$]' %} active{% endif %}\">Completed</a></li>\r\n</ul>"
```
joshuami, line 201: @jookyg , this should be "Construction Menu" I believe.

PR #1487, opened 2020-07-22
POWR-2440 disabled openid_connect in portland module and implemented as submodule

jookyg:


web/modules/custom/portland/modules/portland_openid_connect/src/Routing/RouteSubscriber.php
```diff
@@ -0,0 +1,84 @@
...
+        '@message' => '$_ENV["PANTHEON_ENVIRONMENT"] not set.',
+      ];
+      $this->loggerFactory->get('portland OpenID')
+        ->warning('@message', $variables);
+    }
+
+    // Custom overrides on Pantheon dev, test, and live environments to enable OpenID login form
+    if(
+      isset($_ENV['PANTHEON_ENVIRONMENT']) &&
+      in_array($_ENV['PANTHEON_ENVIRONMENT'], ['powr-2440', 'dev', 'test', 'live'])
```
joshuami, line 63: @jookyg, we still need the demo site to use our openID login.

PR #1518, opened 2020-08-15
POWR-2593 policies in groups POWR-2616 construction clone POWR-2649 migration status

joshuami:
This PR combines several simple stories into a larger PR for easier deployment:

* POWR-2593 Make policies group aware and add group menu link to group policies search view
* POWR-2616 update construction cloning permission
* POWR-2649 add migration status to groups with views to report on status
* POWR-2590 update cirlceCI config to only deploy non-dev dependencies in composer


joshuami:
Had to remove POWR-2590 update cirlceCI config to only deploy non-dev dependencies in composer from the build. I forgot that the `drush cim` relies on a config split that uninstalls several dev packages. Need to think about that one a bit more.

PR #1555, opened 2020-09-14
POWR-2713 Update search_api_* modules

kxwang:


web/sites/default/config/search_api.server.searchstax.yml
```diff
@@ -34,17 +34,17 @@ backend_config:
...
+    core: Test
     timeout: 5
     index_timeout: 5
     optimize_timeout: 10
     finalize_timeout: 30
     commit_within: 1000
-    username: 'null'
-    password: 'null'
+    username: Drupal
+    password: 'Zw92%C*VcxMo'
```
joshuami, line 47: Hmm. I don't think we want a secret in our repo 😳

joshuami (changes requested):

web/sites/default/config/search_api.server.searchstax.yml
kxwang: Hmmm, thank you for catching that. It seems updating the module overwrote the keys setting.

PR #1557, opened 2020-09-14
POWR-2647 Move status message to Login message block

radmosley:


joshuami (changes requested):

web/sites/default/config/block.block.messages.yml
```diff
@@ -1,22 +1,20 @@
...
 status: true
 dependencies:
   module:
     - system
   theme:
     - cloudy
-_core:
-  default_config_hash: 9RnnTUBcn0WRWXfrGOnId2H0Q31uQUeu_F5TnDxl2yg
-id: cloudy_messages
+id: messages
```
joshuami, line 9: When I added back status messages, it was not ideal that this changed as it didn't need to, but it doesn't hurt anything as we don't use the id cloudy_messages in our custom code or theme.

joshuami (approved):

PR #1565, opened 2020-09-18
POWR-2746 media/document permalinks proof of concept work

jookyg:
**This work is being done in branch POWR-2368-1 and the associated multidev environment.**

This branch/PR does the following:

* Creates a group-path token that is used to create group-based files subdirectories
  * _Works with multi-segment group path values such as "omf/toc"_
* Updates file fields in media and content entities to save files in group-based subdirectories
* Updates image fields in media and content entities to save files in subdirectory "[group-path]/img"
* Enables the Rabbit Hole Media (rh_media) submodule
* Configures Rabbit Hole for documents to use the media entity default path as the permalink for the file, for anonymous users only

### DIFF
--- a/web/modules/custom/portland/portland.tokens.inc
+++ b/web/modules/custom/portland/portland.tokens.inc
@@ -51,6 +51,11 @@ function portland_token_info()
     'description' => 'A token to display standardized page titles for Portland.gov',
   ];
 
+  $info['tokens']['portland']['group-shortname'] = [
+    'name' => t('Portland group'),
+    'description' => 'A token to display the group shortname or acronym',
+  ];
+
   return $info;
 }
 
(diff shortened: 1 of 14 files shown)


web/modules/custom/portland/portland.tokens.inc
```diff
@@ -127,6 +132,29 @@ function portland_tokens($type, $tokens, array $data = array(), array $options =
...
+          $route_match = \Drupal::routeMatch();
+          // Entity will be found in the route parameters.
+          if (($route = $route_match->getRouteObject()) && ($parameters = $route->getOption('parameters'))) {
+            // Determine if the current route represents an entity.
+            foreach ($parameters as $name => $options) {
+              if ($name == 'group') {
+                if (isset($options['type']) && strpos($options['type'], 'entity:') === 0) {
+                  $entity = $route_match->getParameter($name);
+                  if ($entity) {
+                    $field = $entity->get('field_shortname_or_acronym');
```
joshuami, line 146: Hmm. There is a catch here. Shortname or Acronym is not required. Maybe we should use field_group_path instead. If we do, can we make it a safe token that adds the extra directory for groups that are within the path of their parent?

web/sites/default/config/core.extension.yml
```diff
@@ -18,6 +18,7 @@ module:
   chosen_lib: 0
   ckeditor: 0
   ckeditor_config: 0
+  clamav: 0
```
joshuami, line 21: I think this might have had a config export done before an initial `lando refresh` that includes a config import. 🤔

web/sites/default/config/field.field.media.chart.image.yml
```diff
@@ -23,7 +23,7 @@ translatable: true
 default_value: {  }
 default_value_callback: ''
 settings:
-  file_directory: '[date:custom:Y]-[date:custom:m]'
+  file_directory: '[portland:group-shortname]'
```
joshuami, line 26: Since this is an image, do we want to append `/img` the same as images?

web/sites/default/config/rabbit_hole.behavior_settings.media_type_document.yml
```diff
@@ -0,0 +1,11 @@
+uuid: 1ba8856f-fe58-4231-ba80-497147ff0dba
+langcode: en
+status: true
+dependencies:
+  config:
+    - media.type.document
+id: media_type_document
+action: page_redirect
```
joshuami, line 8: Did we figure out the way we were going to test whether this redirect breaks analytics? I'd rather we not break analytics first then figure out a way to fix it 😆

joshuami (changes requested):

joshuami:
Updated PR with Jira issue number that will link the issue to the PR.

PR #1577, opened 2020-09-25
POWR-2764 Fix errors during media group reassignment

kxwang:


web/sites/default/monolog.services.yml
```diff
@@ -1,23 +0,0 @@
-parameters:
```
joshuami, line 1: @kxwang, any thoughts on why this yml files was touched?

kxwang: That's an accidental commit. Let me revert it.

joshuami (approved):

PR #1579, opened 2020-09-28
POWR-2746 Document permalinks implementation

jookyg:
The initial proof of concept work was done in branch powr-2368-1. I've merged that branch into a new 2746 branch so that it matches the story ID.

This 2746 branch includes updates requested from [PR 1565](https://github.com/eGovPDX/portlandor/pull/1565):
* Use field_group_path instead of field_shortname_or_acronym for the token that creates group-based file organization
* Images are now placed in an /img subdirectory
* Document field/link templates updated to link to media entity and not directly to the file
* Document Rabbit Hole redirect in effect for both anonymous and authenticated users (sitewide administrators always bypass)


web/sites/default/config/field.field.node.city_code.field_documents_and_exhibits.yml
```diff
@@ -23,7 +23,7 @@ translatable: true
 default_value: {  }
 default_value_callback: ''
 settings:
-  file_directory: code
+  file_directory: citycode
```
joshuami, line 26: The actual directory going forward for code is /code/[title]/[chapter]/[section]. I'm not sure this should be changed to citycode as that was kinda a lame URL pattern from POG 😄

joshuami: I should have marked this comment as part of a review :)

joshuami (changes requested):

web/sites/default/config/field.field.node.city_code.field_documents_and_exhibits.yml
jookyg: Done. This was changed back to "code."

joshuami (approved):

PR #1582, opened 2020-09-28
POWR-2755 change set follow ups

joshuami:


jookyg (approved):
Phew, that's a lot of config and twig to look through. Didn't see anything that looked amiss.

web/themes/custom/cloudy/templates/content/construction_project/node--construction-project.html.twig
```diff
@@ -7,44 +7,34 @@
...
-            </span>
-          </div>
-        </div>
-        {% if content.field_location %}
-        <div class="col-md-6 col-sm-12">
-          {{ content.field_location }}
-        </div>
+      <div class="event__date">
+        <span class="d-flex align-items-center">
+        {% if content.field_display_date|render|striptags|trim is not empty %}
```
brittanysmart, line 12: @joshuami What you have above will always override any set project dates if a value exists in field_display_date, disregarding the checkbox. Use the display date toggle value as the check to determine whether or not to show the display date.

brittanysmart: You could even check the toggle = true and display date is not empty.

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
```diff
@@ -25,20 +25,26 @@
...
+        <a href="https://www.google.com/maps/dir/{{ node.field_location[0].entity.field_address_or_entrance.address_line1 }} {{ node.field_location[0].entity.field_address_or_entrance.address_line2 }} {{ node.field_location[0].entity.field_address_or_entrance.locality }} {{ node.field_location[0].entity.field_address_or_entrance.administrative_area }} {{ node.field_location[0].entity.field_address_or_entrance.postal_code }}" target="_blank">Get Directions</a>
+      </div>
+    {% else %}
+      <div class="get-directions-link">
+        <a href="https://www.google.com/maps/dir/{{ node.field_location[0].entity.field_geolocation.lat }},{{ node.field_location[0].entity.field_geolocation.lon }}" target="_blank">Get Directions</a>
+      </div>
+    {% endif %}
+    {{ content|without('field_location','field_floor','field_emergency_closure') }}
+    <div class="more-information">
+      <a href="{{ url }}" rel="bookmark">More about this location</a>
```
brittanysmart, line 47: "More about this location" isn't being displayed. We should also translate the string.

web/themes/custom/cloudy/templates/field/field--field-end-time.html.twig
```diff
@@ -0,0 +1,17 @@
+{% embed 'field.html.twig' %}
+  {% block content %}
+    {% if multiple %}
+      <div class="field__items">
+    {% endif %}
+    {% for item in items %}
+      <div{{ item.attributes.addClass(item_classes) }}>
+        {{ icon }}
+        <time datetime="{{ item.content }}" class="datetime">{{ item.content }}</time>
```
brittanysmart, line 9: The datetime attribute isn't correctly formatted.

web/themes/custom/cloudy/templates/field/field--field-start-time.html.twig
```diff
@@ -1,6 +1,20 @@
...
-} %}
+{% embed 'field.html.twig' %}
+  {% block content %}
+    {% if multiple %}
+      <div class="field__items">
+    {% endif %}
+    {% for item in items %}
+      <div{{ item.attributes.addClass(item_classes) }}>
+        {{ icon }}
+        <time datetime="{{ item.content }}" class="datetime">{{ item.content }}</time>
```
brittanysmart, line 12: The datetime attribute isn't correctly formatted.

web/themes/custom/cloudy/templates/field/field--field-end-time.html.twig
```diff
@@ -0,0 +1,17 @@
+{% embed 'field.html.twig' %}
+  {% block content %}
+    {% if multiple %}
+      <div class="field__items">
+    {% endif %}
+    {% for item in items %}
+      <div{{ item.attributes.addClass(item_classes) }}>
+        {{ icon }}
```
brittanysmart, line 8: The End Time field shouldn't include an icon.

web/themes/custom/cloudy/templates/field/field-group-html-element.html.twig
```diff
@@ -20,9 +20,7 @@
 {% set classes = classes|default([]) %}
 
 {%
-  set group_classes = [
```
brittanysmart, line 23: The margin variable value is still set in the field group date and time templates. By removing this, we are relying on fields to set margin when it should be the containing element that controls the margin. Also, having unused variables in the group templates is not ideal.

web/themes/custom/cloudy/templates/field/field.html.twig
```diff
@@ -65,24 +65,30 @@
...
 
-{% spaceless %}
+{% apply spaceless %}
   {% block field %}
   <div{{ attributes.addClass(field_classes).addClass(classes) }}>
     {% block label %}
       {% if label_display == "above" %}
         <h2{{ title_attributes.addClass(title_classes) }}>{{ label }}</h2>
-      {% elseif not label_hidden %}
+      {% elseif label_display != "hidden" %}
```
brittanysmart, line 74: I'm curious why you changed this line.

web/themes/custom/cloudy/templates/field/field.html.twig
```diff
@@ -65,24 +65,30 @@
...
         <div{{ title_attributes.addClass(title_classes) }}>{{ label }}</div>
       {%- endif -%}
```
brittanysmart, line 76: You can omit the hyphens as spaceless should remove any whitespace.

web/themes/custom/cloudy/templates/layout/cloudy-main-with-sidebar.html.twig
```diff
@@ -11,12 +11,10 @@
  */
 #}
 <div class="row">
-  <div class="col-lg-8 col-md-12">
+  <div class="cloudy-content col-lg-8 col-md-12">
```
brittanysmart, line 14: Did you add these classes for styling purposes or for clarity?

brittanysmart (changes requested):
Hey @joshuami found a few things to address before merging. Happy to discuss tomorrow.

web/themes/custom/cloudy/templates/content/construction_project/node--construction-project.html.twig
joshuami: That's a good idea. I'll write it up as a follow up since we have the same pattern on the project group type and this wasn't s something I touched on this story.

web/themes/custom/cloudy/templates/content/construction_project/node--construction-project.html.twig
```diff
@@ -7,44 +7,34 @@
...
+          {{ content.field_display_date }}
+        {% else %}
+          {% if content.field_end_date|render|striptags|trim is not empty %}
+            {{ content.field_start_date }}
+            <span class="datetime mx-1" aria-label="{{ 'through'|t }}">&ndash;</span>
```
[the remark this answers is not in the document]
joshuami: Ah yes... the field fixes require that I add a helper class to the span for now. I'd like to look at refactoring that to apply the margin without a helper class. Fix forthcoming in the next commit.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
```diff
@@ -23,21 +23,30 @@
                 {{ content.field_all_day }}
               {% elseif content.field_end_time|render|striptags|trim is not empty %}
```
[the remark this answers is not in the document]
joshuami: I think I have the logic for this a bit more worked out in the next commit. I didn't actually touch the logic, but I agree it was not complete.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
```diff
@@ -23,21 +23,30 @@
...
                 <time datetime="{{ content.field_start_time.0 }}">{{ content.field_start_time }}</time>
-                <span aria-label="{{ 'to'|t }}"> &ndash; </span>
+                <span class="mx-2" aria-label="{{ 'to'|t }}"> &ndash; </span>
                 <time datetime="{{ content.field_end_time.0 }}">{{ content.field_end_time.0 }}</time>
               {% else %}
                 <time datetime="{{ content.field_start_time.0 }}">{{ content.field_start_time }}</time>
```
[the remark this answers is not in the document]
joshuami: The datetime issue is known. I want to follow up on that when we refactor the date and time to use Smart Date. The problem is that the time field module only stores the time, so to get the time in a format I could reformat to a 24 hour clock, I'd have to append a date to the time string in a format that could then be reformatted to the desired time strings for output. That would be a mess ☹️ 

For now, I'm going to live with the slightly incorrect datetime string as it is better than no time element at all and wait for the follow up.

I did indeed call the time element twice, so I've modified the code in the event template to only pull in the field which has the time element now. that is a lot cleaner to read as well.

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
```diff
@@ -25,20 +25,26 @@
...
-    <a href="/emergency-closures">Emergency closure information</a>
-  </div>
-{% else %}
-  {% if node.field_location[0].entity.field_geolocation is empty %}
-    <div class="get-directions-link"><a href="https://www.google.com/maps/dir/{{ node.field_location[0].entity.field_address_or_entrance.address_line1 }} {{ node.field_location[0].entity.field_address_or_entrance.address_line2 }} {{ node.field_location[0].entity.field_address_or_entrance.locality }} {{ node.field_location[0].entity.field_address_or_entrance.administrative_area }} {{ node.field_location[0].entity.field_address_or_entrance.postal_code }}" target="_blank">Get Directions</a></div>
+    </address>
+  {% endspaceless %}
+  {% if node.field_emergency_closure.value == 1 %}
+    <div class="location--closure">
+      <span class="badge badge-danger">Emergency Closure</span><br>
```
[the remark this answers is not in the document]
joshuami: Good find. Added POWR-2774 to do a detailed audit of where translation needs to be added. I'd rather not add it to this story for now as it wasn't in the scope and I already added a ton of extra refactoring.

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
```diff
@@ -25,20 +25,26 @@
...
+      <a href="/emergency-closures">Emergency closure information</a>
+    </div>
   {% else %}
-    <div class="get-directions-link"><a href="https://www.google.com/maps/dir/{{ node.field_location[0].entity.field_geolocation.lat }},{{ node.field_location[0].entity.field_geolocation.lon }}" target="_blank">Get Directions</a></div>
+    {% if node.field_location[0].entity.field_geolocation is empty %}
```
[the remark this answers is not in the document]
joshuami: The address field has its own conditional logic. You have to add the country, which then requires the state, zip, etc. I agree this is ripe for refactor at some point, but it wasn't what I was trying to tackle in this story. I'll add this to the translations refactor as we would be touching every get directions link at that time. 🤔

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
```diff
@@ -25,20 +25,26 @@
...
+      <div class="get-directions-link">
+        <a href="https://www.google.com/maps/dir/{{ node.field_location[0].entity.field_address_or_entrance.address_line1 }} {{ node.field_location[0].entity.field_address_or_entrance.address_line2 }} {{ node.field_location[0].entity.field_address_or_entrance.locality }} {{ node.field_location[0].entity.field_address_or_entrance.administrative_area }} {{ node.field_location[0].entity.field_address_or_entrance.postal_code }}" target="_blank">Get Directions</a>
+      </div>
+    {% else %}
+      <div class="get-directions-link">
+        <a href="https://www.google.com/maps/dir/{{ node.field_location[0].entity.field_geolocation.lat }},{{ node.field_location[0].entity.field_geolocation.lon }}" target="_blank">Get Directions</a>
```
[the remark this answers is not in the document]
joshuami: Believe it or not, this is intentional to match Google's format for directions links. Try going to maps.google.com, searching for an address, then clicking get directions. The resultant URL has `dir//` 🤷  Go figure.

I'll close out @radmosley 's story as works as designed. 😄

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
joshuami: It's not intended to display if a location has the emergency closure triggered. Otherwise this works.

web/themes/custom/cloudy/templates/field/field--field-end-time.html.twig
joshuami: Got it.

web/themes/custom/cloudy/templates/field/field--field-end-time.html.twig
joshuami: See earlier comment about datetime and future refactoring.

web/themes/custom/cloudy/templates/field/field--field-start-time.html.twig
joshuami: See earlier comment about datetime and future refactoring.

web/themes/custom/cloudy/templates/field/field-group-html-element.html.twig
joshuami: Every usage we have has a fields for date and time. After standardizing the field spacing every instance looked a bit weird because of the additional top and bottom spacing on the field group. Hence the removal. I didn't clean up the passing of margin from the group-date, group-project-date, and group-time. I've done that now.

I'm definitely interested in the refactor of all this date logic. Even when we originally rolled this out, I didn't necessarily care for theming the field group to achieve the output. It has a lot of drawbacks. (For instance, that method won't work if we move to Layout Builder.) I couldn't think of a better way to do it at the time, but I'm kinda keen on the Smart Date approach going forward as it would really simplify a lot of this logic. 😄 

I went ahead and removed the margin variables in this next commit.

web/themes/custom/cloudy/templates/field/field.html.twig
joshuami: Is that what those hyphens do?! I had wondered 😄

web/themes/custom/cloudy/templates/field/field.html.twig
joshuami: I think I swapped it because the statement wasn't triggering for me with label_hidden. It should have, but both do the same logic.

web/themes/custom/cloudy/templates/layout/cloudy-main-with-sidebar.html.twig
joshuami: Clarity. We aren't using that layout just yet, but the cloudy-content class is used for targetting TOC. If we are ever able to move to Layout Builder for the default display, this logic would need to be set up this way for header styling as well. 

As a side note, I'm not sure we'll be able to use Layout Builder on any display mode that has if/else logic in it. Layout Builder just doesn't have a means to handle fields that have a dependency on each other because each layout builder block is independent. It makes things like a checkbox to toggle the display or not display of a field kinda impossible. 🙁 

Definitely something to consider as we try to figure out how to incorporate or not incorporate layout builder in future work.

joshuami: Incidentally, I discovered these layout builder limitations when I started to pull in a content type into this layout and realized I couldn't add the logic from the event template. It was definitely a lightbulb moment.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
```diff
@@ -10,9 +10,9 @@
       <div class="row">
         <div class="col mb-4">
           <div class="event__date d-flex">
-            {% if content.field_end_date|render|striptags|trim is not empty %}
+            {% if content.field_start_date.0|render|striptags is not empty and content.field_end_date|render|striptags|trim is not empty %}
```
brittanysmart, line 13: Since Start Date is a required field you could refactor this to only check if Multiple days = true and End Date is not empty; else display start date.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
```diff
@@ -21,23 +21,35 @@
...
-              {% elseif content.field_end_time|render|striptags|trim is not empty %}
-                <time datetime="{{ content.field_start_time.0 }}">{{ content.field_start_time }}</time>
-                <span aria-label="{{ 'to'|t }}"> &ndash; </span>
-                <time datetime="{{ content.field_end_time.0 }}">{{ content.field_end_time.0 }}</time>
+              {% elseif content.field_start_time.0|render|striptags|trim is not empty and content.field_end_time.0|render|striptags|trim is not empty %}
+                {{ content.field_start_time }}
+                <span class="mx-2" aria-label="{{ 'to'|t }}"> &ndash; </span>
+                {{ content.field_end_time }}
+              {% elseif content.field_start_time.0|render|striptags|trim is empty and content.field_end_time.0|render|striptags|trim is not empty %}
+                <span class="mr-1">Ends </span>
```
brittanysmart, line 29: Which content type does "Ends " apply to again?

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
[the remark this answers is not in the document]
brittanysmart: wha??? wow.

web/themes/custom/cloudy/templates/content/service-location/node--service-location--related.html.twig
brittanysmart: Why do we hide it if there's an emergency closure? https://powr-2755-portlandor.pantheonsite.io/service-locations/development-services-center "More about this location" is displayed.

web/themes/custom/cloudy/templates/field/field-group-html-element.html.twig
brittanysmart: Yes, the datetime fields have been overly complex, making theming difficult.

web/themes/custom/cloudy/templates/layout/cloudy-main-with-sidebar.html.twig
brittanysmart: We are living the same life, lol. I just discovered this the other day as well, but I think I have an apporach that could work with layout builder. *fingers-crossed*

brittanysmart (changes requested):

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
joshuami: Ends is for when bids and proposals only put in an end time and no start time. It represents the closing of a big or proposal period.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
joshuami: This works, so let's leave it for now and revisit as a part of Smart Date refactoring.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
joshuami: Odd I can't comment on your comment about "more about this location". 🤔 

I guess I'll just add it here. That secondary more about this location is on the location (building with address) rather than on the service location (place with operational hours). 

That could definitely get a refactor but it is out of the scope for what I was trying to do here. I'll add that to the follow up where we are looking at translation on that field since we'll touch all those more links that that time.

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
```diff
@@ -21,23 +21,35 @@
           <div class="event_time d-flex">
               {% if content.field_all_day|render|striptags|trim is not empty %}
                 {{ content.field_all_day }}
-              {% elseif content.field_end_time|render|striptags|trim is not empty %}
-                <time datetime="{{ content.field_start_time.0 }}">{{ content.field_start_time }}</time>
-                <span aria-label="{{ 'to'|t }}"> &ndash; </span>
-                <time datetime="{{ content.field_end_time.0 }}">{{ content.field_end_time.0 }}</time>
+              {% elseif content.field_start_time.0|render|striptags|trim is not empty and content.field_end_time.0|render|striptags|trim is not empty %}
+                {{ content.field_start_time }}
+                <span class="mx-2" aria-label="{{ 'to'|t }}"> &ndash; </span>
```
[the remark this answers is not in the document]
joshuami: Done!

PR #1598, opened 2020-10-08
POWR-2683 Migrate WaterWorks app to Portland.gov

jookyg:


web/apps/water/waterworks/index.cfm
```diff
@@ -0,0 +1,7 @@
+<cf_pogWrapper site_id="148" c="65997" force_login="no">
```
joshuami, line 1: Why did you include the index.cfm?

web/sites/default/local.services.yml
```diff
@@ -5,7 +5,7 @@
 parameters:
   http.response.debug_cacheability_headers: false
   twig.config:
-    debug: true
+    debug: false
```
joshuami, line 8: We don't want to turn this off on local.

joshuami (changes requested):
More feedback is getting added to the Jira ticket. We need to theme the new time fields added here.

web/sites/default/local.services.yml
jookyg: Oops, I turned that off for my BDS Contacts work so that I could get clean markup in my json feed, but it somehow got saved in this branch. Will fix.

web/themes/custom/cloudy/templates/content/construction_project/node--construction-project.html.twig
```diff
@@ -7,20 +7,32 @@
         {{ content.field_summary.0 }}
       </p>
       {{ content.field_featured_media }}
-      <div class="event__date">
-        <span class="d-flex align-items-center">
-        {% if content.field_display_date|render|striptags|trim is not empty %}
```
joshuami, line 12: While we wanted the time formatting, we still need the conditional display date. I'm adding this back in.

web/themes/custom/cloudy/templates/content/construction_project/node--construction-project.html.twig
```diff
@@ -7,20 +7,32 @@
         {{ content.field_summary.0 }}
```
joshuami, line 7: Also noticed that "summary" is not required, so we had some content with the empty lead paragraph displaying. I'm going to add a temporary conditional statement and a TODO to clean this up by updating the summary field template to have classes that we target rather than explicit class overrides.

joshuami (approved):

PR #1602, opened 2020-10-12
POWR-2396-4 Enable content translation for event, news, alert, and notification

RichardDavies:


web/sites/default/config/core.base_field_override.node.alert.menu_link.yml
```diff
@@ -11,7 +11,7 @@ bundle: alert
 label: 'Menu link'
 description: 'Computed menu link for the node (only available during node saving).'
 required: false
-translatable: false
+translatable: true
```
joshuami, line 14: This one is a little odd. Alerts don't use the menu link field. (It's not even displayed as we never intend someone to go to the default display. 🤔 )

joshuami (changes requested):
One small question to review tomorrow. Otherwise looks good to merge.

web/sites/default/config/core.base_field_override.node.alert.menu_link.yml
RichardDavies: Yeah, I think I was just following the pattern of other entities that do use that field so they have it checked. I guess the question is whether or not having it translatable matters in this case since it's not used.

RichardDavies: Just pushed an amended commit that makes all of the menu link fields non translatable.

joshuami (approved):

PR #1629, opened 2020-10-28
POWR-2860 link field validation

jookyg:
Set link fields for Efiles and webinars to accept external links only, in hopes this may cut down on validation issues for editors.


web/sites/default/config/field.field.node.event.field_web_conference_link.yml
```diff
@@ -23,6 +23,6 @@ translatable: false
 default_value: {  }
 default_value_callback: ''
 settings:
-  link_type: 17
+  link_type: 16
```
joshuami, line 26: Wow... that is an awful way to say "external links only" 😁

joshuami (approved):

PR #1633, opened 2020-11-02
POWR-2915 Add back Behat

kxwang:


tests/features/202_group_content.feature
```diff
@@ -27,6 +27,7 @@ Feature: Members can manage group content
     And I should see "Legacy path"
 
     When I fill in "edit-title-0-value" with "Test page"
+    And I fill in "edit-field-menu-link-text-0-value" with "test-page"
```
joshuami, line 30: Whoops. I put my comment in the commit rather than a review for requested changes. The Short Title field is used for both menu text and the path alias pattern. It should be title case text rather than a path part. We should update this before we merge.

joshuami (changes requested):

tests/features/202_group_content.feature
kxwang: Title updated. Also removed tags on revision logs for the Group Content Revisions Admin view.

joshuami (approved):

PR #1659, opened 2020-11-17
POWR-2974 Remove batch alias generation after group_path update

kxwang:


joshuami (changes requested):
I think we can remove this TODO: `    // <TODO:>LOW PRIORITY: rEGenerate URL aliases for group action links using group path (ie. /bds/content, /bds/members)</TODO:>
`.

Seems like that is covered by extended path aliases now. 

Does the code really need the force group path update? Saving the group doesn't do the trick? I'm just wondering if we could drop that code as well. 🤔

kxwang:
Removed the entire hook. Please review. Please note that content without short title will get the bad URL again after bulk generation.

> I think we can remove this TODO: ` // <TODO:>LOW PRIORITY: rEGenerate URL aliases for group action links using group path (ie. /bds/content, /bds/members)</TODO:>`.
> 
> Seems like that is covered by extended path aliases now.
> 
> Does the code really need the force group path update? Saving the group doesn't do the trick? I'm just wondering if we could drop that code as well. 🤔

joshuami:
Thanks for the update. We should be good to go on all pages and services with short titles. We manually added those for all published content. All unpublished content has the bad paths, but should get fixed as editors try to edit the content and are required to add a short title.

joshuami (dismissed):

joshuami (approved):

PR #1675, opened 2020-12-03
POWR-2422 - Updates to search api database views for group construction 

jookyg:
* Reconfigured facets a bit
* Added toggle checkbox field to enable past construction for groups


web/sites/default/config/core.entity_view_display.group.program.default.yml
```diff
@@ -227,7 +229,6 @@ hidden:
   field_enable_programs_menu_item: true
   field_enable_projects_menu_item: true
   field_enable_public_notices: true
-  field_enable_reports: true
```
joshuami, line 230: @jookyg , this is a weird one. For some reason this change unhid the field_enable_reports. We don't want that showing up at the bottom of program pages. Likely just need to move it back to hidden here from /admin/group/types/manage/program/display and then cex -y.

joshuami (changes requested):

joshuami (approved):

PR #1677, opened 2020-12-03
POWR-2376 Set translation permissions

RichardDavies:


web/sites/default/config/group.role.bureau_office-admin.yml
```diff
@@ -93,9 +96,26 @@ permissions:
   - 'delete own group_node:service_location content'
   - 'edit group'
   - 'leave group'
+  - 'translate any group_media:audio entity'
```
joshuami, line 99: Wouldn't all group types need these permissions set? This looks like only bureau_office was updated.

composer.json
```diff
@@ -197,7 +197,8 @@
                 "2457999 - Cannot use relationship for rendered entity on Views": "https://www.drupal.org/files/issues/2020-06-17/cannot-use-relationship-2457999-244.patch",
                 "2998826 - Term route context (allows block placed on term pages)": "https://www.drupal.org/files/issues/2019-04-12/term_route_context-2998826-31.patch",
                 "3114467 - 'Negate' form value for condition plugins should be cast to boolean in validation (for Google Tag Manager)": "https://www.drupal.org/files/issues/2020-03-23/3114467-8.patch",
-                "962664 - Taxonomy Index for unpublished entities": "https://www.drupal.org/files/issues/2020-04-16/962664-drupal-add-unpublished-nodes-to-taxonomy-index-149.patch"
+                "962664 - Taxonomy Index for unpublished entities": "https://www.drupal.org/files/issues/2020-04-16/962664-drupal-add-unpublished-nodes-to-taxonomy-index-149.patch",
+                "2972308 - Allow users to translate content they can edit": "https://www.drupal.org/files/issues/2020-07-02/2972308-49.patch"
```
joshuami, line 201: Given that this is in Drupal 9.1, adding this patch is pretty low risk. Nice find.

joshuami (changes requested):
I think this just needs to have the other group types updated with the permissions and we should be good to go.

web/sites/default/config/group.role.bureau_office-admin.yml
RichardDavies: Sorry, those permissions are misleading. I tried another patch that created group-level translation permissions but it didn't offer anything that wasn't already offered by the existing patch I added, so I ended up removing that patch. But it appears that the permissions I created with it were still exported to the config. TLDR: those group permissions don't currently do anything the universal "translate content I can edit" permission allows the user to edit content in any group type.

joshuami (approved):

PR #1686, opened 2020-12-11
POWR-2981 Council document follow ups

joshuami:


web/modules/custom/portland/portland.module
```diff
@@ -915,6 +915,13 @@ function _notify_reviewer($entity, $entity_event_type)
         'draft' => 'Draft',
         'archived' => 'Unpublished/Archived',
         'published' => 'Published',
+        'unpublished' => 'Unpublished',
```
kxwang, line 918: If this is user facing, will need to remove "Unpublished" from the display message for "archived" to avoid confusion.

kxwang (approved):

web/modules/custom/portland/portland.module
joshuami: Good catch. The "archived" machine name is associated with the "Unpublished" label in the default workflow. The label of "Unpublished/archived" is legacy. 😄

PR #1688, opened 2020-12-14
POWR-3058 Delete all related multi-devs when a branch is merged

kxwang:


.circleci/config.yml
```diff
@@ -140,14 +140,18 @@ build_and_deploy: &build_and_deploy
...
+            if [[ $lowercase_last_log =~ ^(powr-[0-9]+)[[:space:]]+.+$ ]]; then
+              echo "Found the branch name: '${BASH_REMATCH[1]}'";
               # Check if a multidev environment with the branch name exists or not
               terminus multidev:list --format=list --field=Name portlandor > ./env_list.txt
-              if grep -Fxq "${BASH_REMATCH[2]}" ./env_list.txt; then
-                # Delete both the multidev site and branch
-                terminus multidev:delete --delete-branch portlandor.${BASH_REMATCH[2]} -y
-              fi
+              # Pattern to match "powr-1234a" or "powr-1234-1". Not "powr-12345"
+              pattern="^${BASH_REMATCH[1]}([a-z]*|(-\d+))\s$" && echo $pattern
```
RichardDavies, line 150: I think we should also match the pattern powr-1234-a where 'a' can be 1 or more letters.

.circleci/config.yml
```diff
@@ -140,14 +140,18 @@ build_and_deploy: &build_and_deploy
             # Example: Merge pull request #35 from eGovPDX/powr-152
             cd /var/www/html
             # Must read the last commit message from Github. The last one in local repo is "CircleCI Deployment"
-            if [[ $(git log origin/master --pretty=format:%s -1) =~ (^Merge pull request #[0-9]* from eGovPDX\/)([^,]*) ]]; then
-              echo "Found a Pull Request Merge for branch: '${BASH_REMATCH[2]}'";
+            last_log=$(git log origin/master --pretty=format:%s -1)
+            lowercase_last_log=`php -r "print strtolower('$last_log');"`
+            if [[ $lowercase_last_log =~ ^(powr-[0-9]+)[[:space:]]+.+$ ]]; then
+              echo "Found the branch name: '${BASH_REMATCH[1]}'";
               # Check if a multidev environment with the branch name exists or not
```
RichardDavies, line 147: I think the term "branch name" is slightly misleading here. Aren't we really referring to a story ID here? I think we should be intentional and consistent about using the right terminology here since a branch name and story ID are sometime, but now always the same.

RichardDavies (changes requested):
Overall, I think this looks really good. There are a few small changes though I think we should make.

.circleci/config.yml
kxwang: Great feedback! Pattern updated accordingly.

.circleci/config.yml
kxwang: Change the comment to "story ID".

.circleci/config.yml
```diff
@@ -140,14 +140,18 @@ build_and_deploy: &build_and_deploy
             # Example: Merge pull request #35 from eGovPDX/powr-152
             cd /var/www/html
             # Must read the last commit message from Github. The last one in local repo is "CircleCI Deployment"
-            if [[ $(git log origin/master --pretty=format:%s -1) =~ (^Merge pull request #[0-9]* from eGovPDX\/)([^,]*) ]]; then
-              echo "Found a Pull Request Merge for branch: '${BASH_REMATCH[2]}'";
+            last_log=$(git log origin/master --pretty=format:%s -1)
+            lowercase_last_log=`php -r "print strtolower('$last_log');"`
+            if [[ $lowercase_last_log =~ ^(powr-[0-9]+)[[:space:]]+.+$ ]]; then
+              echo "Found the story ID: '${BASH_REMATCH[1]}'";
               # Check if a multidev environment with the branch name exists or not
```
RichardDavies, line 147: "branch name" could be changed to story ID here as well.

.circleci/config.yml
```diff
@@ -140,14 +140,18 @@ build_and_deploy: &build_and_deploy
...
               terminus multidev:list --format=list --field=Name portlandor > ./env_list.txt
-              if grep -Fxq "${BASH_REMATCH[2]}" ./env_list.txt; then
-                # Delete both the multidev site and branch
-                terminus multidev:delete --delete-branch portlandor.${BASH_REMATCH[2]} -y
-              fi
+              # Pattern to match "powr-1234a" or "powr-1234-1". Not "powr-12345"
```
RichardDavies, line 149: Could add the new pattern powr-1234-a to your comment here.

RichardDavies (approved):
I know this is getting really nit-picky, but here are a few related changes that could be made to the comments for consistency. But given that they're just comments, I'll let you decide if you want to make those changes or not.

RichardDavies (approved):
Looks prefect!! :)

joshuami (approved):

PR #1694, opened 2020-12-18
POWR-2904 Base multidevs off of the PR's base branch

RichardDavies:


.circleci/config.yml
```diff
@@ -173,51 +173,60 @@ build_and_deploy: &build_and_deploy
...
+              #   - Extract PR number from pull request URL
+              #   - Query GitHub API for PR data and extract base branch from JSON base.ref property
+              [[ $CIRCLE_PULL_REQUEST =~ \/pull\/([0-9]+)$ ]]
+              BASE=$(curl -s https://api.github.com/repos/egovPDX/$CIRCLE_PROJECT_REPONAME/pulls/${BASH_REMATCH[1]} | grep -Pzo '(?s)"base":.*?"ref": "\K[A-Za-z0-9_-]+' | tr -d '\0')
+              if [[ $BASE = "master" ]]; then
+                BASE=dev
+              fi
+              # Create a new Multi-Dev with files and database from site corresponding to PR's base branch.
+              echo "Creating new multidev site cloned from $BASE. Please wait." 
+              terminus multidev:create $CIRCLE_PROJECT_REPONAME.$BASE $PANTHEON_ENV
```
kxwang, line 190: This probably is a corner condition: If the base branch multi-dev was deleted manually, the script can fall back on Dev since that's guaranteed to be there. Base branch multi-dev can be checked by running "grep -Fxq "$CIRCLE_PROJECT_REPONAME.$BASE" ./env_list.txt;"

RichardDavies: I like your out-of-the-box thinking, but if the base branch has been deleted then that automatically means the PR is inoperative since the code can no longer be merged into that selected base branch. So there's no point in accommodating this edge case since you can't merge the code anyway. If that case ever happens, I think it's fine for the build to fail. At that point, the PR would need to be updated with a valid base branch which would trigger a new build that should then succeed.

RichardDavies:
I realized this isn't the PR I want merged. This one was just to test using translate as the base branch. Please review and merge #1695 instead as that one merges into master.

PR #1720, opened 2021-01-20
POWR-3150 Fix group permit search

kxwang:


joshuami:
Included in POWR-3100.

PR #1721, opened 2021-01-21
POWR-3150 Convert Group Services/Permits pages to Solr search

kxwang:


joshuami:
We may follow up on converting back to Solr search at a future date. I think it will likely be our best choice after Pantheon has a search service for us to use that it co-hosted with the rest of their platform.

PR #1738, opened 2021-02-03
POWR-3142 add auditor_blog and auditor_blog_redirect to migration module

radmosley:


web/modules/custom/portland_migrations/migrations/auditor_blog.yml
```diff
@@ -0,0 +1,38 @@
+id: auditor_blog
+label: 'Auditor Blog Import'
+migration_group: auditor
```
RichardDavies, line 3: The migration_group is used to link together all migrations that are related to each other and should be run together. So for this migration, it should probably be set to 'auditor_blog'.

web/modules/custom/portland_migrations/migrations/auditor_blog_redirects.yml
```diff
@@ -0,0 +1,34 @@
+id: audit_blog_redirects
+label: 'Migrate Auditor Blog legacy paths into Redirects table'
+langcode: en
+migration_group: auditor
```
RichardDavies, line 4: The migration_group is used to link together all migrations that are related to each other and should be run together. So for this migration, it should probably be set to 'auditor_blog'.

web/modules/custom/portland_migrations/migrations/auditor_blog.yml
```diff
@@ -0,0 +1,38 @@
...
+  header_row_count: 1
+  keys:
+    - CONTENT_ID
+process:
+  uid:
+    plugin: default_value
+    default_value: 1
+  title: CONTENT_NAME
+  field_summary: SUMMARY_TEXT
+  field_body_content/value: TEXT
```
RichardDavies, line 16: In order to migrate embedded body content images into media items, this field should use the `migrate_body_content_and_linked_media` plugin. Here's an example from the pbot_news migration:
```
field_body_content/value:
    plugin: migrate_body_content_and_linked_media
    source: TEXT
    group_id: 14
```
You'll just need to change the group_id to the ID of the auditor's group so it knows where to save the image media entities it creates.

RichardDavies (changes requested):
I suggested a few minor changes to improve a couple of things. Also, I noticed the PR includes several file changes unrelated to the migration... I'm not sure why but it should probably only contain the readme, the .csv, and your .yml migration configuration files.

web/modules/custom/portland_migrations/migrations/auditor_blog.yml
```diff
@@ -0,0 +1,40 @@
+id: auditor_blog
+label: 'Auditor Blog Import'
+migration_group: auditor_news_blog
```
RichardDavies, line 3: Why did you change this to `auditor_news_blog` when the other one is `auditor_blog`? Sorry if I confused you, but the migration_group should be the same for both of the .yml files since they both belong to the same group (or set) of migrations that need to be run together. Grouping them together allows you to run all of the migrations in the group with a single command instead of running each migration separately.

Technically, the group name can be anything, as long as it's the same value for all files in the group, but for consistency with our other migrations, we've typically named the group the same value as the primary migration file of the group, which would be `auditor_blog` in this case. Then you can run both migrations in the group using `lando terminus drush [environment] -- migrate:import --group=auditor_blog`

RichardDavies (changes requested):
Sorry, I think I may have confused you about the migration_group value.

joshuami (approved):

PR #1743, opened 2021-02-09
POWR-3175 council document follow ups and toolbar_menu implementation

joshuami:


web/themes/custom/cloudy/src/css/_overrides.scss
```diff
@@ -261,3 +261,7 @@ drupal-entity {
 .block-portland-login-block {
   margin-top: $cloudy-space-4;
 }
+
+// Override print variables
+$print-page-size:                   letter !important;
+$print-body-min-width:              map-get($grid-breakpoints, "sm") !important;
```
brittanysmart, line 267: We set up bootstrap's breakpoints to match our token values
You can replace the value with the token value. That way it will use the topmost override.
`$print-body-min-width: $cloudy-breakpoint-sm;`

brittanysmart (changes requested):
Only one minor change, looking good otherwise.

web/themes/custom/cloudy/src/css/_overrides.scss
joshuami: I'll replace this with our custom token—though it was copied directly from the _variables.scss way of approaching this type of override. I still think the way Bootstrap sets up is easier to read and understand than the tokens approach, but I will concede for now until we can refactor our approaches to be more of a Drupal/Bootstrap override and less custom.

brittanysmart: I was thinking follow the current pattern until we can refactor the apporach so we know where to look. Make it easier to find and replace.

PR #1747, opened 2021-02-11
POWR-3161 search help text update

ccdatapdx:


web/themes/custom/cloudy/templates/search/search-api-page.html.twig
```diff
@@ -11,6 +11,7 @@
   {% else %}
     {{ no_results_found }}
   {% endif %}
+  To improve search results, be specific. Use quotes to search for exact text.
```
joshuami, line 14: @ccdatapdx , I altered the text a bit to make it work a bit better with the results statement. I also moved it below the if/else clause so that it is only printed once. (Don't repeat yourself if you don't have to 😉 )

joshuami (approved):

PR #1748, opened 2021-02-12
POWR-3144 Mega menu button style

RichardDavies:


web/themes/custom/cloudy/src/components/_header.scss
```diff
@@ -258,7 +286,8 @@
   // Content region - /front, /search, /search/*, /404
   &--large {
     border-bottom: $search-large-border-style-underline;
-    border-radius: $cloudy-space-0 $cloudy-space-0 $border-radius-sm $border-radius-sm;
+    border-radius: $cloudy-space-0 $cloudy-space-0 $border-radius-sm
+      $border-radius-sm;
```
joshuami, line 290: @RichardDavies , do you know why these values are breaking to the next line? Code lint?

web/themes/custom/cloudy/templates/layout/page.html.twig
```diff
@@ -98,7 +94,10 @@
 
       {# menu toggle #}
       <button{{ toggle_attributes|without('class').setAttribute('data-target', '#' ~ menu_id).setAttribute('aria-controls', menu_id).setAttribute('aria-label', 'toggle site menu'|t).addClass(['cloudy-header__toggle', 'cloudy-header__toggle--menu']) }}>
-        <span class="toggle__label">{{ 'menu'|t|capitalize }}</span>
+        <span class="toggle__label">
+          <span class="icon toggle-icon icon-menu"></span>
+          {{ 'menu'|t|capitalize }}
```
joshuami, line 99: Hmm. Just noticed this because of the reformatting. I believe that all caps is  going to read as M - E - N - U to screen readers. We should capitalize this with CSS rather than a twig transformation.

RichardDavies: The capitalize filter only capitalizes the first letter, turning it into "Menu". That wasn't my code so I'm not sure why they're using that filter instead of just changing the string from 'menu' to 'Menu'.

web/themes/custom/cloudy/src/components/_header.scss
RichardDavies: Hmm, I thought I'd fixed that in my last story. I'm not sure if Brit re-introduced it with her commit or if it is some code linting. I've just pushed up a new commit to restore it to one line.

web/themes/custom/cloudy/templates/layout/page.html.twig
joshuami: That's weird. That probably a couple of CPU cycles wasted. I think it should just be {{ 'Menu'|t }} to match the way Drupal core usually handles this. I suppose you could argue that if every string were entered as lowercase with capitalize it would allow you to translate the string to appear both as lower or capitalized, but you could apply the same by doing {{ 'Menu'|t|lower }} in the few spots you want lowercase. 

In general, I would think it would be better to match the pattern in core for this sort of thing. Doesn't need to block this PR, but probably something we should refactor to be only one approach in our templates going forward. 🤔

RichardDavies: It's a quick easy change if you want me to remove the filter and just capitalize the string.

RichardDavies: It just occurred to me that the purpose of the capitalize function is mostly likely because the string is being translated and string translations are case sensitive. This allows us to have a single translation for the word "menu" and capitalize it when desired.

joshuami: That's kinda what I was referring to in my last comment. I don't have a strong feeling about it one way or another, but Drupal core seems to take the approach of passing the capitalized string in to translation rather than a lowercase string. You could always lower the case on output for the few places that would be needed—or transform those places with CSS. 

Drupal defaults to sentence case in nearly all of the UI. After a little further reading, running strings through capitalize has its drawbacks. The biggest is when you have UI elements that include abbreviations like "RSS feed". I think this is why they settled on sentence case for default labels so long ago.

Not a blocker for this PR, but maybe something to hash out as a team standard—ideally defaulting to Drupal standards when they apply.

joshuami (approved):

PR #1753, opened 2021-02-19
POWR-3247 Refactor: Standardize theme z-index values

brittanysmart:


web/themes/custom/cloudy/src/components/_skip-link.scss
```diff
@@ -8,7 +8,7 @@
     -webkit-transform: translateX(-50%);
     -ms-transform: translateX(-50%);
     transform: translateX(-50%);
-    z-index: 50;
+    z-index: 1;
```
RichardDavies, line 11: I'm curious why this skip link, which is visually hidden, has all of these styles to alter it's positioning, appearance, and z-index (none of which seem to matter since it's visually hidden...) This isn't really a blocker, just something I noticed that could potentially be removed if unnecessary to reduce our CSS size.

RichardDavies (approved):

web/themes/custom/cloudy/src/components/_skip-link.scss
brittanysmart: I had the same thought. The skip link does appear when focused on by tabbing to it or using other assistive technology. Therefore, we need to provide styling – OR we could remove our custom skip link style and instead use Drupal core skip link styling.

I wonder if our current styling is problematic as the skip link overlaps our alert text (when present), which could impact readability/intrusion on the visual design. @julegan did you want to provide any further thoughts/recommendations on our current usage?

Resources;
http://web-accessibility.carnegiemuseums.org/code/skip-link/
https://webaim.org/techniques/skipnav/

Our custom skip link style:
Screen Shot 2021-03-01 at 10 21 34 AM


Drupal core skip link style:
Screen Shot 2021-03-01 at 10 22 28 AM


City of Boston skip link:
Screen Shot 2021-03-01 at 10 24 53 AM

RichardDavies: Ah, that makes sense. I guess the code would be more self-explanatory if those styles were applied to the :focus pseudo selector instead of the main element though. But it works either way.

PR #1756, opened 2021-02-23
POWR-3186 Liquor License Applications view

RichardDavies:


web/sites/default/config/views.view.db_search_group_documents.yml
```diff
@@ -444,3 +447,280 @@ display:
...
+        area_text_custom:
+          id: area_text_custom
+          table: views
+          field: area_text_custom
+          relationship: none
+          group_type: group
+          admin_label: ''
+          empty: false
+          tokenize: false
+          content: "<p>You may submit comments about a liquor license application within 30 days of the application’s receipt by the City. They will become part of the permanent public record for that application, available upon request.</p>\r\n\r\n<p>All comments must include the address of the liquor license application as well as your own name and address, or they will not be considered. To submit comments, email them to <a href=\"mailto:liquor@portlandoregon.gov\">liquor@portlandoregon.gov</a>; or send them by postal mail to:</p>\r\n\r\n<p class=\"mx-auto w-50\">Liquor Licensing<br />\r\n1120 SW 5th Ave, Suite 114<br />\r\nPortland, OR 97204</p>"
```
joshuami, line 701: @RichardDavies , we should probably use the address element and let it left justify for the mailing address. I can't think of another place where we center content in this fashion.

web/sites/default/config/facets.facet.group_document_liquor_application_neighborhood.yml
```diff
@@ -0,0 +1,61 @@
...
+    show_numbers: true
+    soft_limit: 0
+    soft_limit_settings:
+      show_less_label: 'Show less'
+      show_more_label: 'Show more'
+    show_reset_link: true
+    reset_text: 'Show all'
+    hide_reset_when_no_selection: true
+query_operator: and
+use_hierarchy: true
```
joshuami, line 30: QA did not test the use of a child neighborhood to make sure the hierarchy correctly displays. When I reviewed, I could not immediately get a child neighborhood to render the city section correctly. Perhaps it was because there was only one result for that neighborhood?

As a side note, the entity reference autocomplete field isn't ideal for selecting a neighborhood. It requires that the person entering the data know all the neighborhoods by exact name. We don't need to fix that in this PR, but I think I will make sure to add a select2 widget to this field when we roll out the feature to the editors in Civic Life.

joshuami (changes requested):

web/sites/default/config/views.view.db_search_group_documents.yml
RichardDavies: I've made this change. I didn't realize the HTML 5 spec now allows the address element to be used for general purpose addresses (it used to be just for the document author's contact info.)

web/sites/default/config/facets.facet.group_document_liquor_application_neighborhood.yml
RichardDavies: I'm not exactly sure what you mean by a child neighborhood, but perhaps this is something we can resolve when we refactor this view in the following story.

joshuami: I must mean that the terms in that vocabulary are arranged in a hierarchy. Editors will usually want to select the deepest term they can, which is usually a specific neighborhood. (Sextant > Coalition > Neighborhood) If you use the neighborhood facets on the construction view, you'll see what I mean.

It think it would be fine to resolve in that story.

joshuami (approved):

PR #1758, opened 2021-02-24
POWR-3145 Search form widget changes

RichardDavies:


web/themes/custom/cloudy/cloudy.theme
```diff
@@ -114,13 +114,17 @@ function cloudy_form_search_api_page_block_form_search_portland_gov_alter(&$form
   /* merge returned search class */
   $form['#attributes']['class'] = array_merge($form['#attributes']['class'], $additional_classes);
 
-  $text = t("Search Portland.gov");
+  if ($is_front_page) {
+    $text = t("Search Portland.gov");
+  } else {
+    $text = t("Search");
```
joshuami, line 120: I think we might want to  keep this as "Search portland.gov" for the rest of the site as well. Bureaus think of themselves as islands and the sitewide search pages like News and Events have a second "search" form. Making it clear that search is all of Portland.gov makes more sense in the navigation context. 🤔

web/themes/custom/cloudy/src/components/_header.scss
```diff
@@ -227,101 +228,131 @@
...
       }
     }
   }
 }
 
+
 // TODO: Override module library with libraries-override
 // Search API block autocomplete
-.ui-autocomplete {
+.ui-autocomplete.search-api-autocomplete-search {
```
joshuami, line 351: .ui-autocomplete is also used by the entity reference field autocomplete. This change breaks those fields.

joshuami, line 351: .ui-autocomplete is also used by the entity reference field autocomplete. This change breaks those fields.

RichardDavies: Ah, so it appears that our CSS for the header Search API autocomplete fields was bleeding over into all autocomplete fields (perhaps this was intentional, but it is certainly contrary to what the CSS comments and filename indicate these styles are for.) My change limited the CSS to only the Search API autocomplete fields which left the entity reference fields with their default styles. So I guess I need to restore some basic styles for the non Search API autocomplete fields.

joshuami (approved):

PR #1770, opened 2021-03-04
POWR-3212-2 adding lang attribute to allowed text formats

ccdatapdx:


web/sites/default/config/filter.format.simple_editor.yml
```diff
@@ -16,7 +16,7 @@ filters:
...
     weight: -10
     settings:
-      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <ol type start> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p> <hr> <sup> <sub> <s> <a href hreflang !href accesskey id rel target title> <i class=""> <br> <drupal-entity data-entity-type data-entity-uuid data-entity-embed-display data-entity-embed-display-settings data-align data-caption data-embed-button data-langcode alt title> <table class> <caption> <tbody> <thead class> <tfoot class> <th class> <td class> <tr class> <abbr title> <caption class>'
+      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <ul type> <p lang=""> <div lang=""> <span lang=""> <a lang=""> <ol type start> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p> <hr> <sup> <sub> <s> <a href hreflang !href accesskey id rel target title> <i class=""> <br> <drupal-entity data-entity-type data-entity-uuid data-entity-embed-display data-entity-embed-display-settings data-align data-caption data-embed-button data-langcode alt title> <table class> <caption> <tbody> <thead class> <tfoot class> <th class> <td class> <tr class> <abbr title> <caption class>'
```
joshuami, line 19: The `a` attribute was duplicated in your change. Also, looks like the only change we need to make is to add `span` as lang is allowed without explicitly declaring it.

web/sites/default/config/filter.format.simplified_editor_with_media_embed.yml
```diff
@@ -17,7 +17,7 @@ filters:
...
+      allowed_html: '<em> <strong> <cite> <blockquote cite> <code> <p lang=""> <div lang=""> <span lang=""> <a lang=""> <ul type> <li> <dl> <dt> <dd> <h2 id> <h3 id> <h4 id> <h5 id> <h6 id> <p class=""> <hr> <sup> <sub> <s> <img src alt data-entity-type data-entity-uuid> <pre> <i class=""> <drupal-entity data-* title alt data-entity-type data-entity-uuid data-entity-embed-display data-entity-embed-display-settings data-align data-caption data-embed-button data-langcode> <a href hreflang !href accesskey id rel target title class=" btn-cta btn-cta-outline"> <div class> <ol type start class="code-section-list"> <br> <table class summary> <th scope colspan rowspan> <tr class scope> <td class scope colspan rowspan> <thead class> <tbody class> <tfoot class> <abbr title> <caption class> <u>'
```
joshuami, line 20: See other comment.

joshuami (changes requested):

web/sites/default/config/filter.format.simple_editor.yml
joshuami: Oh... you duplicated the `p` element as well. I didn't see if `div` was duplicated. 🤔

ccdatapdx: Ah ok, I didn't realize that. Thanks for the flag, will fix.

joshuami (approved):

PR #1792, opened 2021-03-19
POWR-3315 Move database search based views to Solr

kxwang:


joshuami:
Closing for a rewrite of the approach.

PR #1808, opened 2021-03-30
POWR-2699 total results added to group content view and delete old gr…

ccdatapdx:
…oup content view


joshuami (changes requested):
On second thought... I just realized that the view we are removing is actually created by the Group module on installation. Let's update the view description with a note that the view is created by the Group module and is not currently in use on Portland.gov and add "DEPRECATED" to the title instead of removing the view.

PR #1811, opened 2021-04-01
POWR-3263 Add 311 field to Contact

kxwang:


web/sites/default/config/field.field.node.contact.field_display_311.yml
```diff
@@ -0,0 +1,29 @@
...
+third_party_settings:
+  custom_add_another:
+    custom_add_another: ''
+    custom_remove: ''
+id: node.contact.field_display_311
+field_name: field_display_311
+entity_type: node
+bundle: contact
+label: 'Display 311'
+description: 'Mark as True to display 311 info.'
```
joshuami, line 19: Added revised help text and permission check to the Jira issue.

joshuami (changes requested):

web/sites/default/config/field.field.node.contact.field_display_311.yml
kxwang: Fixed in new commit.

joshuami (approved):

PR #1815, opened 2021-04-05
POWR-1267 use Claro for all entity creation forms

joshuami:
This PR makes the following changes:

* Adds select2 for use instead of Chosen on entity forms
* Updates the group module to add an advanced sidebar to group entity forms
* Turns Chosen off for admin forms (we'll need a follow up to remove from any places in use for the public theme)
* Makes the default media browser view work with Claro
* Updates every single entity form display to work with Claro 😅 
* Updates some miscellaneous help text to improve forms with new details layout


joshuami:
Closing in favor of the PR to release branch.

PR #1824, opened 2021-04-13
POWR-3414 updates to circleci and lando scripts for timing.

joshuami:


kxwang (approved):

web/private/scripts/drush_post_deploy_maintenance_on.php
```diff
@@ -0,0 +1,12 @@
...
+
+// Do NOT run for Dependabot branches
+if (preg_match('/^bot-\d+/', $_ENV['PANTHEON_ENVIRONMENT'])) {
+    die('Dependabot branch detected. Aborting!');
+}
+
+
+// Turn on maintenance mode.
+echo "Turn on maintenance mode...\n";
+system('drush drush sset system.maintenance_mode TRUE 2>&1');
```
jookyg: Is the redundant drush call here intentional?

web/private/scripts/drush_post_deploy_maintenance_off.php
```diff
@@ -0,0 +1,12 @@
...
+
+// Do NOT run for Dependabot branches
+if (preg_match('/^bot-\d+/', $_ENV['PANTHEON_ENVIRONMENT'])) {
+    die('Dependabot branch detected. Aborting!');
+}
+
+
+// Turn off maintenance mode.
+echo "Turn off maintenance mode...\n";
+system('drush drush sset system.maintenance_mode FALSE 2>&1');
```
jookyg: Is the redundant drush call here intentional?

jookyg (approved):

web/private/scripts/drush_post_deploy_config_import.php
```diff
@@ -8,5 +8,5 @@
 
 // Import all config changes.
 echo "Start importing configuration from yml files...\n";
-system('drush config-import -y 2>&1');
+system('drush config-import -y -vvv 2>&1');
```
RichardDavies, line 11: Do you think it would be helpful to output a timestamp before and after the config import? (So that we'll know the exact time it starts and finishes, might be helpful if we need to correlate with external events.) 

Something like `system('date +"%T"');` should do the trick.

RichardDavies (commented):
Looks good, but I do have one possible suggestion.

web/private/scripts/drush_post_deploy_config_import.php
joshuami: That's a great idea. I've added and pushed. Could you give it another review when the build finishes?

PR #1829, opened 2021-04-15
POWR-3404 Rename Manage: Group: views and convert to Claro

kxwang:


tests/behat.yml
```diff
@@ -30,7 +30,7 @@ default:
       region_map:
         header: "div#header"
         main_menu: "#block-cloudy-main-menu"
-        tabs: "div#tabs"
+        tabs: "ul.tabs"
```
joshuami, line 33: @kxwang , please resolve the merge conflict with this line. In an earlier story that was merged, we had to updated the tabs to use the Claro tabs, this may work, but I think the claro_tabs region that was created for tests might be a little more targeted. Let me know what you think.

joshuami (changes requested):

joshuami (approved):

PR #1833, opened 2021-04-15
POWR-3411 Migrate Behat tests to Jest

kxwang:


.lando.yml
```diff
@@ -18,15 +18,12 @@ services:
       - update-ca-certificates -f -v
     run_as_root:
       - /bin/sh -c "if [ ! -d /var/www/.terminus/plugins/terminus-secrets-plugin ]; then { mkdir -p /var/www/.terminus/plugins; composer create-project -d /var/www/.terminus/plugins pantheon-systems/terminus-secrets-plugin:~1; chown -R www-data:www-data /var/www/.terminus; } fi"
+      - /bin/sh -c "ln -sf /usr/local/bin/composer /app/vendor/bin/composer"
```
kxwang, line 21: @RichardDavies @joshuami I added this line to ensure Composer is V2 after rebuild/restart.

RichardDavies: Looks good to me. 

Although looking at this now, I wonder why some of our Lando commands use `/bin/sh -c` and others are just directly invoked... I _think_ `/bin/sh -c` is only necessary for things like `if then` statements which are internal bash statements. Everything else that is an external command like `sed`, `cp`, and `ln` could probably just be directly invoked. Although there's no real harm to invoking a separate instance of the shell for these commands beyond maybe a very small performance hit and just sacrificing some simplicity/readability in our `.lando.yml` file. Anyway, just an observation for a possible minor refactor either now or in the future.

kxwang: You are right! I've removed the /bin/sh part.

PR #1844, opened 2021-04-22
POWR-3437 hotfix to correct draw map feature on construction

joshuami:


web/sites/default/config/views.view.group_geo_map.yml
```diff
@@ -278,7 +278,7 @@ display:
           group_type: group
           admin_label: ''
           label: ''
-          exclude: false
+          exclude: true
```
joshuami, line 281: @kevin, I made this small change to make this EVA match the one for the construction geo_map EVA. It doesn't seem to resolve the issue that the map doesn't display for the project group type.

kxwang (approved):

PR #1850, opened 2021-04-23
POWR-3439 cleanups from after Sprint 60 releases

joshuami:


web/private/scripts/slack_deploy_notification.php
```diff
@@ -162,12 +162,11 @@ function _get_secrets($requiredKeys, $defaults)
 /**
  * Send a notification to slack
  */
-function _slack_notification($slack_url, $channel, $username, $text, /*$attachment,*/ $alwaysShowText = true)
+function _slack_notification($slack_url, $username, $text, /*$attachment,*/ $alwaysShowText = true)
 {
   // $attachment['fallback'] = $text;
   $post = array(
     'username' => $username,
-    'channel' => $channel,
```
RichardDavies, line 170: Why didn't you remove username here as well? If you're not going to pass in that parameter you'll need to remove it here too. Are you sure this function with work without the username and channel values?

RichardDavies (changes requested):
I'm not sure your changes to the _slack_notification function parameters are all valid.

web/private/scripts/slack_deploy_notification.php
```diff
@@ -162,12 +162,11 @@ function _get_secrets($requiredKeys, $defaults)
 /**
  * Send a notification to slack
  */
-function _slack_notification($slack_url, $channel, $username, $text, /*$attachment,*/ $alwaysShowText = true)
+function _slack_notification($slack_url, $username, $text, /*$attachment,*/ $alwaysShowText = true)
```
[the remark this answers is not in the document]
joshuami: It does pass the $username still. The $user name is a default variable near the start of the file.

web/private/scripts/slack_deploy_notification.php
joshuami: Okay... trying again with a focus on only touching the channel.

RichardDavies (approved):
Ok, I think it looks good now.

PR #1858, opened 2021-04-27
POWR-3425 Use terminus build:workflow:wait to wait for code sync event

kxwang:


.circleci/config.yml
```diff
@@ -113,147 +125,108 @@ build_and_deploy: &build_and_deploy
...
+          echo "Waiting for the workflow: Sync code on $PANTHEON_ENV..."
+          terminus build:workflow:wait --max 240 $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV "Sync code on \"$PANTHEON_ENV\""
+
+          echo "Rebuilding cache..."
+          terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- cr
+          echo "Running database updates..."
+          terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- updb -y
+          echo "Importing config..."
+          terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- cim -y -vvv
+          echo "Running cron..."
```
joshuami, line 170: I've removed the cron run on our deploys to test and prod because it often conflicted with the every 5 minute cron run on prod. Do we really need to run cron? `drush deploy` didn't include it so I felt comfortable removing.

.circleci/config.yml
```diff
@@ -113,147 +125,108 @@ build_and_deploy: &build_and_deploy
...
+              JIRA_STORY_ID=${BASH_REMATCH[2]}
+            # Squash and Merge on other feature branches
+            # Example: POWR-3058 Delete all related multi-devs when a branch is merged
+            elif [[ $lowercase_last_log =~ ^(powr-[0-9]+)[[:space:]]+.+$ ]]; then
+              echo "Found story ID from squash and merge commit message: '${BASH_REMATCH[1]}'";
+              JIRA_STORY_ID=${BASH_REMATCH[1]}
             else
-              echo "export PANTHEON_ENV=$CIRCLE_BRANCH" >> $BASH_ENV && source $BASH_ENV
+              echo "Cannot find JIRA story ID from the last commit message: $lowercase_last_log"
+              exit 1
```
joshuami, line 196: Am I correct in thinking this will fail our build if there isn't a multidev to delete?

joshuami: Will that cause issues (build fail) when we merge the `release` branch into `master`

joshuami: Nevermind... found the if `dev` or `release` at the start of this code 😄

.circleci/config.yml
```diff
@@ -297,7 +270,7 @@ run_tests: &run_tests
           - "/tmp/vendor"
     - restore_cache:
         name: Restore Behat cache
-        key: v2-behat-dependencies-{{ checksum "tests/composer.lock" }}
+        key: v2-behat-dependencies-{{ checksum "tests/composer.lock" }}
```
joshuami, line 273: That's odd... why did it add a space?

joshuami (approved):
Approving, but I did have a couple of comments from review that could be doublechecked.

I agree a second review from @RichardDavies would be nice 😄

.circleci/config.yml
kxwang: Cron removed

.circleci/config.yml
kxwang: Space removed

joshuami (approved):

.circleci/config.yml
```diff
@@ -113,147 +125,106 @@ build_and_deploy: &build_and_deploy
...
-            terminus connection:set $CIRCLE_PROJECT_REPONAME.dev git
-            git push -f pantheon master
-            # TODO: Sanitize DB here
-            # terminus env:clone-content $CIRCLE_PROJECT_REPONAME.test dev
+          else
+            echo "export PANTHEON_ENV=$CIRCLE_BRANCH" >> $BASH_ENV && source $BASH_ENV
+          fi
+
+          # Check if the multidev branch site exists. If not, create it
+          terminus multidev:list --format=list --field=Name $CIRCLE_PROJECT_REPONAME > ./env_list.txt
```
RichardDavies, line 139: This will try to create a multidev site named "dev" for commits to master now that this code block executes for all builds. I think this code block for checking for existing multidevs and creating them still needs to be within the "else" that only runs for commits that aren't on master.

RichardDavies (changes requested):
I think I found a couple of things that should be changed, one trivial and the other pretty significant. (Sorry the code review took so long, I got hung up yesterday trying to troubleshoot my lando issue.)

.circleci/config.yml
```diff
@@ -96,14 +99,23 @@ build_and_deploy: &build_and_deploy
...
           # Don't check for host key for Pantheon sites
           echo -e "host *.drush.in\n    StrictHostKeyChecking no" >> ~/.ssh/config
+
           # Terminus log in with the user email that will be displayed as the commit user in Pantheon Git repo
           terminus auth:login --machine-token="$eGov_Terminus" --email="$eGov_Email"
+          
           # Git operations
+          # Set the git commit user name and email address so it's clear in Pantheon log who made the change
+          git config --global user.name $CIRCLE_USERNAME
+          git config --global user.email $eGov_Email
```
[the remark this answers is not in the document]
kxwang: Removed the extra user email config

.circleci/config.yml
kxwang: Great catch! Moved the site creation logic into "not master" block.

.circleci/config.yml
```diff
@@ -214,46 +150,80 @@ build_and_deploy: &build_and_deploy
...
-            echo "Importing config..."
-            terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- cim -y -vvv
-            echo "Rebuilding cache"
-            terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- cr
+          # Set the connection type to GIT and force push
+          terminus connection:set $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV git
+          git push -f pantheon $CIRCLE_BRANCH
+          # Wait upto 240 seconds for the Sync code workflow to finish
+          echo "Waiting for the workflow: Sync code on $PANTHEON_ENV..."
+          terminus build:workflow:wait --max 240 $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV "Sync code on \"$PANTHEON_ENV\""
```
RichardDavies, line 161: I'm curious, why the max wait of 240 seconds? I would think we would want to always wait until it's finished. If the wait ever times out and we proceed before the event has finished, it could lead to build failure or other issue.

RichardDavies (dismissed):
I approve, although I do have one question about timing out our sync code wait after 240 seconds.

.circleci/config.yml
kxwang: You are right that the code would continue after the timeout. The ideal way is to error out when the timeout happens but I can't seems to figure out how to do that with the build plugin. Without setting timeout here, there must be some other timeout settings that will terminate the job eventually but I'm not sure where they are.

RichardDavies: Well, my recommendation would be to remove the `--max 240` timeout from the command so that it will wait indefinitely for the event to finish. Should we ever encounter an issue where the Pantheon event doesn't finish for some reason and causes the build to be stuck waiting, we can manually stop the CircleCI job. This should be a rare, if ever occurrence, but it seems better to deal with it that way than to risk a build failure if an event takes 300 seconds to run for some reason.

.circleci/config.yml
```diff
@@ -158,7 +158,7 @@ build_and_deploy: &build_and_deploy
           git push -f pantheon $CIRCLE_BRANCH
           # Wait upto 240 seconds for the Sync code workflow to finish
```
RichardDavies, line 159: I know now I'm getting picky, but the comment didn't get updated. :)

RichardDavies (approved):

joshuami (approved):

PR #1862, opened 2021-05-03
POWR-3374 Webform to submit neighborhood association events

RichardDavies:


.lando.yml
```diff
@@ -13,9 +13,10 @@ services:
     build:
       - "/app/vendor/bin/phpcs --config-set installed_paths /app/vendor/drupal/coder/coder_sniffer"
     build_as_root:
+      - sed -i 's/DEFAULT@SECLEVEL=2/DEFAULT@SECLEVEL=1/' /etc/ssl/openssl.cnf
```
joshuami, line 16: @RichardDavies , do we need this in the PR if we can no longer use lando on the network? I would be okay with removing this from our lando build steps as a quick refactor, but adding changes that don't affect the ability to use lando is probably not necessary. Thoughts?

joshuami (approved):

.lando.yml
RichardDavies: Yeah, this new line should be unnecessary now that I'm outside of the network decryption. I've just removed it but I left the existing cert stuff as I feel like that should be removed as part of a separate story where it can be properly tested.

PR #1866, opened 2021-05-05
POWR-3407 add content access checks to search api processors

joshuami:


kxwang (approved):
Looks good. Not sure how it works exactly but wonder if the user will see a perceivable performance hit.

PR #1870, opened 2021-05-10
POWR-3482 Create contact type taxonomy

RichardDavies:


web/sites/default/config/language.content_settings.taxonomy_term.contact_types.yml
```diff
@@ -0,0 +1,11 @@
+uuid: f878f2b0-c98e-4051-b111-d61b398c8a86
+langcode: en
+status: true
+dependencies:
+  config:
+    - taxonomy.vocabulary.contact_types
```
joshuami, line 6: The machine names should really be singular rather than plural. I missed this on construction_types back in the day ☹️ (All the other vocabularies look correct.)

joshuami (changes requested):

joshuami (approved):

PR #1871, opened 2021-05-10
POWR-3460 Certified Advisory Group custom output needs to be moved to a field template

kxwang:


web/sites/default/config/field.field.group.advisory_group.field_certified_advisory_body.yml
```diff
@@ -25,5 +25,5 @@ default_value:
 default_value_callback: ''
 settings:
   on_label: '<a href="/civic/advisorygroups">Certified Advisory Body</a>'
-  off_label: 'Off'
+  off_label: '<span></span>'
```
joshuami, line 28: Interesting solution. 😄

joshuami (approved):

PR #1896, opened 2021-05-25
POWR-3540 remove requirement for contact address and hide by default

joshuami:


web/sites/default/config/core.entity_form_display.node.contact.default.yml
```diff
@@ -80,7 +81,23 @@ content:
...
+      conditional_fields:
+        c686403d-4f9c-499f-9f1f-1f9420b9e1b5:
+          dependee: field_contact_type
+          settings:
+            state: visible
+            condition: value
+            grouping: AND
+            values_set: 3
+            value: ''
+            values: "619\r\n620\r\n621\r\n622\r\n623\r\n624\r\n625"
```
RichardDavies, line 94: I just added a new term "City Required Notice Contact" to this taxonomy on production since we decided to also migrate that contact type from the ONI DB. It's id is 628 if you'd like to add it to the list now, otherwise we'll have to add it later.

RichardDavies (approved):
Looks good, although I have a suggestion to avoid a future edit.

web/sites/default/config/core.entity_form_display.node.contact.default.yml
joshuami: Good idea. Fortunately, I still had that branch open on my local 😄

RichardDavies (approved):

PR #1928, opened 2021-06-10
POWR-3223 Neighborhood search result and teaser display modes

RichardDavies:


web/sites/default/config/core.entity_view_mode.taxonomy_term.teaser.yml
```diff
@@ -0,0 +1,14 @@
+uuid: cd08ae6a-534c-481b-a79b-e890f49ed4ac
+langcode: en
+status: true
+dependencies:
+  module:
+    - lightning_core
```
joshuami, line 6: Hmm. I wonder why it has a dependency to lightning core. 🤔

web/themes/custom/cloudy/templates/taxonomy/taxonomy-term.html.twig
```diff
@@ -28,6 +28,8 @@
   {% set vocabulary = term.bundle|join %}
   {% if 'community_actions' in vocabulary %}
     {% set vocabulary = 'service type'|t|title %}
+  {% elseif vocabulary == 'neighborhood' %}
+    {% set vocabulary = content.field_neighborhood_type %}
```
joshuami, line 32: Hmm. I feel like this is going to include the field template and include some divs we don't want. You should be able to call something like `content.field_neighborhood_type.0.value` to get a string with only the value and nothing else. We can do this in the follow up story where it is easier to review.

joshuami (approved):

joshuami:
Neither issue is enough to fail the story, but please write up some follow ups.

PR #1949, opened 2021-06-29
POWR-3570 Change neighborhood fields to select2

radmosley:


web/sites/default/config/core.entity_form_display.group.project.default.yml
```diff
@@ -533,6 +534,7 @@ content:
           options: ''
           popup: 0
         control: 0
+      path: '{"color":"#3388ff","opacity":"1.0","stroke":true,"weight":3,"fill":"depends","fillColor":"*","fillOpacity":"0.2","radius":"6"}'
```
joshuami, line 537: @radmosley , do you know where this change is coming from? It doesn't seem like a mapping integration should be happening in this change. 🤔

joshuami (commented):
I think this is safe to approve. The change to the map field on the construction and project forms doesn't seem to affect anything. It's probably the first time that config has been updated since an upstream change to the leaflet module or the like.

joshuami (approved):

PR #1955, opened 2021-07-03
POWR-3631 Bureau/Office homepages

brittanysmart:
-  Adds back-reference view to display a group's parent and elected official


web/sites/default/config/clamav.settings.yml
```diff
@@ -0,0 +1,15 @@
+enabled: true
```
joshuami, line 1: @brittanysmart , this should not be here. It means you did a `lando cex` after your `lando latest` and before you did your `lando refresh`. That caused the local config to get overridden with the dev config.

web/sites/default/config/core.extension.yml
```diff
@@ -12,10 +12,13 @@ module:
   block: 0
   block_content: 0
   block_exclude_pages: 0
+  bootstrap_layout_builder: 0
+  bootstrap_styles: 0
   breakpoint: 0
   bulk_user_registration: 0
   ckeditor: 0
   ckeditor_config: 0
+  clamav: 0
```
joshuami, line 21: clamav, pantheon_advanced_cache, and redis lines should be removed from this file. It is part of the local config getting overridden by dev.

web/sites/default/config/search_api.index.documents.yml
```diff
@@ -147,6 +147,7 @@ processor_settings:
   entity_status: {  }
   hierarchy:
     fields:
+      field_display_groups: group-field_bureaus
```
joshuami, line 150: I'm not certain this should have changed. You aren't touching documents in this story and the documents search index doesn't have any specific group relationships that should have been impacted by any of the other changes.

web/themes/custom/cloudy/src/css/_overrides.scss
```diff
@@ -283,3 +283,8 @@ drupal-entity {
 // Override print variables
 $print-page-size:                   letter !important;
 $print-body-min-width:              $cloudy-breakpoint-sm;
+
+
+.more-link{
+  text-align: left;
+}
```
joshuami, line 290: Needs line break.

web/themes/custom/cloudy/templates/field/field--group--label--bureau-office--default.html.twig
```diff
@@ -0,0 +1,7 @@
+{% include '@components/heading/heading.twig' with {
+    url: url,
+    level: level,
+    style: style_override,
+    text: element[0],
+    classes: ['order-1', 'mb-2']
+} only %}
```
joshuami, line 7: Needs line break.

web/themes/custom/cloudy/templates/views/views-view-fields--field-set--block-1.html.twig
```diff
@@ -0,0 +1,72 @@
+{#
+/**
+ * @file
+ * Theme override to display all the fields in a row.
```
joshuami, line 4: I'm not certain what this view template is overriding in the current implementation. All the variables seem to be date related. If we are going to keep this, we should at least update the notes and perhaps rename the block_1 machine name to something with a bit more meaning. 🤔

web/sites/default/config/block.block.views_block__group_notifications_block_1.yml
```diff
@@ -29,3 +30,13 @@ visibility:
     pages: "/group/*\r\n/group/*/*"
     negate: false
     context_mapping: {  }
+  'entity_bundle:group':
```
joshuami, line 33: Could you talk me through why we would need both the page visibility and the entity bundle setting? It seems like it should be one or the other.

joshuami (changes requested):

web/sites/default/config/clamav.settings.yml
brittanysmart: Should I update to false?

web/themes/custom/cloudy/templates/views/views-view-fields--field-set--block-1.html.twig
brittanysmart: I'm removing this view, added by mistake.

web/sites/default/config/search_api.index.documents.yml
brittanysmart: I think wires got crossed when moving previous work. I've removed `field_display_groups`.

web/sites/default/config/core.extension.yml
brittanysmart: clamav, pantheon_advanced_cache, and redis removed.

web/sites/default/config/clamav.settings.yml
joshuami: The best way to fix this order of operations error is to run `git checkout release web/sites/default/config/clamav.settings.yml` to revert the change. Editing any Drupal generated YML file is bad practice. It introduces the possibility of configuration corruption. I try to avoid it when possible. 😄

joshuami: I should note, that when you do that, I believe it will delete this file as clamav settings only exist in the dev config split and up.

web/sites/default/config/core.extension.yml
joshuami: Similar to the clamav settings, it is best not to edit these files directly. However, I can't think of a way to cleanly fix this file through configuration changes. You'll need to test the heck out of both your local and multidev to make sure the manual edit doesn't have unintended consequences.

web/sites/default/config/search_api.index.documents.yml
joshuami: Similar to the previous comments. You should only do this within the UI, then save, then run a `drush cex` to get the change.

web/sites/default/config/clamav.settings.yml
brittanysmart: Thanks for including the command, but it did not work because this file doesn't exist or isn't tracked in git. I ended up deleted this file and committing.

web/sites/default/config/block.block.views_block__group_notifications_block_1.yml
brittanysmart: This addition was part of previous exploration. I've removed it in my last commit.

web/sites/default/config/search_api.index.documents.yml
joshuami: Please don't mark comments as resolved if the task is to remove the file. It will automatically resolve the comment when the file is removed. It's too easy to miss these important tasks if the comment is resolved.

joshuami:
@jookyg and @kxwang, can the two of you please give this PR to initialize the `group-home` branch a look and weigh in on the feedback I wrote into the POWR-3631 story's last comment. 

There is no requirement that we merge this new feature branch, but it should give us a place to explore layout builder a bit more with the possibility of brining it into the release branch at a future date.

jookyg (approved):
I don't see anything out of place or concerning in the PR, though my knowledge of layout builder is sill pretty raw. This looks good for a merge to group-home, but I have reservations about the approach overall, which I noted in a comment in the story. Basically echoing your concerns, Josh.

kxwang (commented):
Added some general comments in Jira for record keeping.

joshuami (approved):

PR #1956, opened 2021-07-07
POWR-3433 Neighborhood & coalition boundary map

RichardDavies:


joshuami (dismissed):
@RichardDavies, apologies that I didn't catch this in my earlier review, but as I was giving this one last look, I noticed a view that had a default machine name that could be updated for clarity. I'm not sure it is enough to fail the PR, but it feels like we won't clean it up if we don't address them as we see them 😄

web/sites/default/config/views.view.neighborhoods.yml
```diff
@@ -1403,6 +1408,350 @@ display:
         - url
         - user.permissions
       tags: {  }
+  entity_view_map:
```
RichardDavies, line 1411: This is the view's machine name, right?

web/sites/default/config/core.entity_view_display.taxonomy_term.coalition.default.yml
```diff
@@ -133,6 +134,12 @@ content:
     third_party_settings: {  }
     type: link
     region: content
+  neighborhoods_entity_view_1:
```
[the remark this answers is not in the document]
RichardDavies: This is the id assigned by layout builder to the EVA, right? I'm not even sure that's editable. The EVA's machine name in the view has already been set to `entity_view_map`.

web/sites/default/config/core.entity_view_display.taxonomy_term.coalition.full.yml
```diff
@@ -375,6 +386,17 @@ content:
     third_party_settings: {  }
     type: link
     region: content
+  neighborhoods_entity_view_1:
```
joshuami, line 389: @RichardDavies , I'm not sure the entity_view_1 is the one assigned by layout builder. Here in the coalitions "full" view mode it is showing both neighborhoods_entity_view_1 and neighborhoods_entity_view_map. Though, they are the same weight 🤔

RichardDavies: Ugh. So when I originally changed the view's machine name, it appears Drupal did not update the references to the original machine name in all of the various config files. I've manually cleaned them up now and removed references to the original machine name.

joshuami (approved):

PR #1965, opened 2021-07-16
POWR-3629 Remove group memberships

RichardDavies:


joshuami (approved):
Hmm. I probably would not have combined the membership management view into the "my groups" view. I don't know if another developer would think to look for it there. That said, it is pretty easy to edit a view with the contextual menu (gears on a view page) and it saves an extra config file.

Approving this to merge.

PR #1987, opened 2021-08-17
POWR-3693 Update group manage content view

radmosley:


web/modules/custom/portland/portland.links.task.yml
```diff
@@ -20,24 +20,6 @@ portland.group.nodes.all:
   weight: 10
 # Define the rest of tab menu item
 # MUST have base_route as the group page route and the parent_id as the tab menu route
-portland.group.nodes.events:
```
joshuami, line 23: @radmosley, please see the Jira story for some more context, but we need to keep these menu items and associated views. They have value to editors. When you push that change, this comment will become outdated, so no need to resolve as it helps me see that the requested change was made.

joshuami (changes requested):

PR #1988, opened 2021-08-17
POWR-3678 Block users based on AD info

kxwang:


web/modules/custom/portland/portland.module
```diff
@@ -770,6 +770,11 @@ function portland_form_alter(&$form, &$form_state, $form_id)
     if( $form_id === "media_document_add_form" || $form_id === "media_document_edit_form" ) {
         $form['#validate'][] = '_validate_document';
     }
+
+    // POWR-3657 - disable HTML 5 validation for all forms
```
joshuami, line 774: Darnit! I accidentally merged @jookyg 's PR into master instead of release, but what's odd is that the change is now showing up here. Do you normally branch off of master or release? I had thought we were all branching off of release. (It shouldn't matter too much, and it helped me find an issue with release being out of date here, but we might want to discuss so we are all on the same page.)

joshuami (approved):

web/modules/custom/portland/portland.module
kxwang: IIRC, I always branched off Release. Not sure why this commit got in there.

PR #1989, opened 2021-08-17
POWR-3649 Group social media migration prework

brittanysmart:


web/sites/default/config/core.entity_form_display.group.bureau_office.default.yml
```diff
@@ -33,12 +33,12 @@ dependencies:
     - field.field.group.bureau_office.field_group_menu_toggle
     - field.field.group.bureau_office.field_group_path
     - field.field.group.bureau_office.field_instagram
-    - field.field.group.bureau_office.field_linkedin
```
joshuami, line 36: @brittanysmart, please check the issue for some quick changes to make sure we don't lose content. After deploying those changes, this comment should show as outdated so you don't need to resolve the comment.

joshuami (changes requested):

web/themes/custom/cloudy/templates/field/field--field-linkedin.html.twig
```diff
@@ -1,18 +0,0 @@
-{% embed 'field.html.twig' with {
```
joshuami, line 1: @brittanysmart, shouldn't we keep these templates for the deprecated field until we have updated the data? If we delete them, the output will break from the current expected layout even if the value is actually there. That will cause the usage to look broken until Oden can get the audit complete. We can make part of his audit the removal of the data in the field so that when complete we can remove the field definitions and these templates in a separate PR. Sound good?

joshuami (changes requested):

web/themes/custom/cloudy/templates/field/field--field-linkedin.html.twig
brittanysmart: AFAIK the deprecated YouTube, LinkedIn, and Nextdoor group fields are not displayed which is why I renamed the template to use our newly added fields. Group homepage templates display either a contact (uses contact item component) or Facebook, Twitter, Instagram fields.

Is there a place where the deprecated fields are currently displayed that I've overlooked?

```
{% set marquee_secondary_right %}
  {% if content.field_contact.0 %}
    {{ content.field_contact }}
  {% else %}
    {{ content.field_email }}
    {{ content.field_phone }}
    {{ content.field_facebook }}
    {{ content.field_twitter }}
    {{ content.field_instagram }}
  {% endif %}
{% endset %}
```

PR #1990, opened 2021-08-17
CSM-65 Follow-up updates

jookyg:


web/modules/custom/portland/modules/portland_location_picker/src/Controller/VerifyLocationController.php
```diff
@@ -0,0 +1,22 @@
...
+   * @return JsonResponse
+   */
+  public function GetMatches($address) {
+    $results = new PortlandMaps();
+    $results = $this->getAddressMatches($address);
+    return new JsonResponse([ 'data' => $results]);
+  }
+
+
+}
```
joshuami, line 22: Technically needs a line break, but we can pick that up in the next PR.

joshuami (approved):

PR #2002, opened 2021-08-26
POWR-3498 Add elected official, group parent, group children to bureau/office homepages

brittanysmart:


web/sites/default/config/core.entity_view_display.group.advisory_group.featured.yml
```diff
@@ -44,7 +44,7 @@ content:
     weight: 0
     label: hidden
     settings:
-      view_mode: banner
+      view_mode: featured
```
joshuami, line 47: Nice catch. That is currently delivering a much bigger image that needed.

joshuami (approved):
Approving and merging. Could you please clean up the old group-home multidev and delete that branch that will not get used.

PR #2011, opened 2021-09-01
POWR-3737 Regression testing, remove notifications from feature branch test.

radmosley:


tests/percy/lib/util.js
```diff
@@ -15,6 +15,16 @@ exports.unmasquerade = async function(page, HOME_PAGE) {
   await page.goto(`${HOME_PAGE}${link}`);
 }
 
+exports.removeAlert = async function(url, HOME_PAGE, page) {
```
kxwang, line 18: An optional improvement is to move "page.goto" outside "removeAlert" to make the intention clear and the code first visits a page and then removes alert from that page. After that, "removeAlert" will only need one parameter "page". This is along the the UNIX way of "small and focused components" :)

tests/percy/lib/util.js
```diff
@@ -15,6 +15,16 @@ exports.unmasquerade = async function(page, HOME_PAGE) {
...
+  await page.goto(`${HOME_PAGE}/${url}`, { waitUntil: "load" });
+    // Remove sitewide alert
+      await page.evaluate(() =>
```
kxwang, line 21: I wonder if an exception will be thrown when there is no alert to begin with. Could you test by unpublishing all alerts? If there is an exception, can add a try/catch block around it. In this case can even ignore the exception since no alert is not an error condition.

tests/percy/lib/util.js
radmosley: Makes sense, changed removeAlert and added the goto back to anonymous.

tests/percy/lib/util.js
radmosley: Tested this and the test still pass if there isn't an alert.

kxwang (approved):

joshuami:
@radmosley , if this is not something we are merging, could you please close this PR and delete the multidev. Thanks!

PR #2020, opened 2021-09-07
POWR-3718 Webform to submit neighborhood contact updates

RichardDavies:


web/sites/default/config/lightning_core.versions.yml
```diff
@@ -248,6 +248,8 @@ select2: 1.13.0
 sendgrid_integration: 1.2.0
 serialization: 0.0.0
 shortcut: 0.0.0
+simple_block: 1.3.0
```
joshuami, line 251: @RichardDavies , this file needs to be reverted. (Though I'm not really sure when lighting_core enable modules actually fires.)

joshuami (changes requested):

web/sites/default/config/lightning_core.versions.yml
RichardDavies: Removed.

joshuami (approved):

PR #2042, opened 2021-09-29
POWR-1694a initial round of updates for Drupal 9 upgrade.

joshuami:


jookyg (approved):
Looks good, though these two files are missing newlines at the end. I don't know if that's an issue for built js and css files.

web/themes/custom/cloudy/dist/bootstrap.bundle.js

PR #2045, opened 2021-09-30
CSM-99 add report e-scooter webform

jookyg:
Included in this PR:

* Report an E-scooter form
* Report an E-scooter Resolution form
* Updated/new contrib modules and libraries
  * drupal/sendgrid_integration: ^2.0
  * mmucklo/inflect: ^0.3.0 (dependency of zendesk_api_client_php)
  * zendesk/zendesk_api_client_php: ^2.2
* Tweaks to Location Picker widget
* New portland_zendesk custom module, adapted from webform_zendesk contrib module
* Configure Zendesk web token to use Key module
* Configure mail system to use Sendgrid Integration for all webforms in all environments (this allows testing without manual config updates in multidev without potentially spamming editors with content review notifications)
* Enable prepopulation of 2 Website Feedback form fields, so that Zendesk api error messages can set form values in link
* Updated standard webform template to include Ticket Type and Agent Summary fields, part of the standards I'm trying to establish and get documented.


web/sites/default/config/webform.webform._template_standard_webform.yml
```diff
@@ -631,7 +644,7 @@ handlers:
     settings:
       states:
         - completed
-      to_mail: gregoryscottclapp@gmail.com
+      to_mail: powr@portlandoregon.gov
```
joshuami, line 647: I wonder if we want to make this 311@portlandoregon.gov in a follow up so that that's our "default". 🤔

joshuami (approved):

web/sites/default/config/webform.webform._template_standard_webform.yml
jookyg: Oh, good catch. I'll update it in my Location widget bug fix story.

PR #2051, opened 2021-10-02
POWR-3741 Create single display mode to display Elected Official on site menu

brittanysmart:


web/sites/default/config/core.entity_view_display.group.elected_official.default.yml
```diff
@@ -64,14 +64,14 @@ third_party_settings:
         - field_person_image
       meta:
         - field_summary
+        - field_contact
```
joshuami, line 67: I'll leave a note in the Jira issue, but we need to lead contact in the contact region to keep the fix we did for electeds with more than one contact.

joshuami (changes requested):

joshuami:
@RichardDavies , could you give this PR a second set of eyes. I like the tile refactor work, but I don't know the cloudy-global-menu implementation quite as well as you do.

web/sites/default/config/core.entity_view_display.group.elected_official.default.yml
brittanysmart: Good catch. Just added the fix.

RichardDavies (approved):
Looks good. I don't see any issues.

joshuami (approved):

PR #2086, opened 2021-10-20
POWR-3527-3 config of struc sync replacing fixed blocks

ccdatapdx:


web/sites/default/config/block.block.views_block__db_search_group_news_block_featured_blog.yml
```diff
@@ -0,0 +1,42 @@
+uuid: 3432d519-ce0e-42ed-9f07-64a4b3677ebf
+langcode: en
```
joshuami, line 2: @ccdatapdx @RichardDavies , this file should not be added back in. I think there may be an issue with the order of git operations.

composer.lock
```diff
@@ -22140,7 +22140,6 @@
     "aliases": [],
     "minimum-stability": "dev",
     "stability-flags": {
-        "drupal/better_exposed_filters": 10,
```
joshuami, line 22143: Hmm. I'm not sure this lock file should have changed either. 🤔

joshuami (changes requested):

composer.lock
RichardDavies: I think this one is okay. I noticed my translation branch has the same change after doing some `composer requires`. so this appears to be expected.

web/sites/default/config/block.block.views_block__db_search_group_news_block_featured_blog.yml
ccdatapdx: removed this file

PR #2104, opened 2021-11-03
POWR-3846 Add languages and enable translations

RichardDavies:


joshuami (approved):
I reviewed every single file 😉

PR #2110, opened 2021-11-05
POWR-3796 Bureau/Office Header and Contacts redesign

brittanysmart:


web/sites/default/config/image.style.group_logo.yml
```diff
@@ -1,26 +1,15 @@
...
+    uuid: 397126e3-0887-4089-a10f-b7eca78a65ce
     id: image_scale
     weight: 2
     data:
-      width: 80
-      height: 40
+      width: null
+      height: 72
       upscale: false
-  542b2d8a-9dc6-49d1-aa6e-f1484b7014a2:
```
joshuami, line 18: The current approach strips out the PNG optimization code that Greg added to all our image styles. That will lead to PNGs that look super fuzzy when uploaded as a PNG.

joshuami (changes requested):

web/sites/default/config/image.style.16_9_800w.yml
```diff
@@ -0,0 +1,17 @@
...
+status: true
+dependencies:
+  module:
+    - focal_point
+name: 16_9_800w
+label: 16_9_800w
+effects:
+  c8babe8f-10c5-47bf-b72c-e5bb1f367845:
+    uuid: c8babe8f-10c5-47bf-b72c-e5bb1f367845
+    id: focal_point_scale_and_crop
```
joshuami, line 12: The change also seems to remove our responsive image style for the featured image/media and makes it a fixed size of 800x450 that gets sized only by CSS.
We talk a lot about how we try and deliver the best sized image based on device type. This may be okay if it somehow gives us retina-grade output, but it seems like a regression if we don't describe the change in the implementation details of the story.

joshuami (changes requested):

web/sites/default/config/image.style.group_logo.yml
brittanysmart: After discussion we decided to discard implementing bitmap-based group logos in favor of SVGs. Some of the benefits to SVGs include pixel-perfect scaling, smaller file sizes, ability to modify styling of SVG using CSS if desired.
Part of this work is already captured in Jira with plans to groom.

web/sites/default/config/image.style.16_9_800w.yml
brittanysmart: Reverted this change in branch [powr-3796c](https://github.com/eGovPDX/portlandor/pull/2166)

PR #2113, opened 2021-11-09
POWR-3844 update group module and apply workflow and permissions

joshuami:


web/sites/default/config/views.view.admin_groups.yml
```diff
@@ -101,22 +101,28 @@ display:
...
-            field_email: field_email
-            field_facebook: field_facebook
-            field_instagram: field_instagram
-            field_twitter: field_twitter
-            field_linkedin_link: field_linkedin_link
-            field_nextdoor_link: field_nextdoor_link
-            field_youtube_link: field_youtube_link
-            field_contact: field_contact
-            edit_group: edit_group
+            field_migration_status: field_migration_status
```
jookyg, line 108: This list of columns changed dramatically with totally different kinds of data. Is that expected?

jookyg (dismissed):
I don't see anything troubling except the list of columns in views.view.admin_groups.yml. The field list is a dramatic change. Is that expected? If so, then this all looks good.

PR #2119, opened 2021-11-17
POWR-3527-4 reconfigurating to keep 404 page manager

ccdatapdx:


joshuami (approved):
@ccdatapdx , in future stories that touch block layout, use the show row weights feature to minimize the amount of reordering of block weights. A lot of these files did not need to change and it makes for a more difficult PR review.

PR #2139, opened 2021-12-07
POWR-3935 Add Spanish and enable translation modules

RichardDavies:


web/themes/custom/cloudy/dist/bootstrap.bundle.js
```diff
@@ -23,5 +23,5 @@
...
...this._element).trigger(e),e.isDefaultPrevented()||this._close()}}},{key:"dispose",value:function(){this._clearTimeout(),this._element.classList.contains("show")&&this._element.classList.remove("show"),i()(this._element).off(cn),i.a.removeData(this._element,"bs.toast"),this._element=null,this._config=null}},{key:"_getConfig",value:function(e){return e=on(on(on({},mn),i()(this._element).data()),"object"===tn(e)&&e?e:{}),s.typeCheckConfig("toast",e,this.constructor.DefaultType),e}},{key:"_setListeners",value:function(){var e=this;i()(this._element).on(cn,'[data-dismiss="toast"]',(function(){return e.hide()}))}},{key:"_close",value:function(){var e=this,t=function(){e._element.classList.add("hide"),i()(e._element).trigger(fn)};if(this._element.classList.remove("show"),this._config.animation){var n=s.getTransitionDurationFromElement(this._element);i()(this._element).one(s.TRANSITION_END,t).emulateTransitionEnd(n)}else t()}},{key:"_clearTimeout",value:function(){clearTimeout(this._timeout),this._timeout=null}}])&&an(t.prototype,n),o&&an(t,o),e}();i.a.fn.toast=gn._jQueryInterface,i.a.fn.toast.Constructor=gn,i.a.fn.toast.noConflict=function(){return i.a.fn.toast=ln,gn._jQueryInterface}}]);
```
joshuami, line 26: It doesn't seem like the js and css assets should have changed in this PR. 🤔

joshuami (changes requested):
Everything looks good except the js and css assets being in the change. That doesn't make sense for just enabling the modules and a `lando refresh` shouldn't change them randomly.

web/themes/custom/cloudy/dist/bootstrap.bundle.js
RichardDavies: Yep, I thought the same thing. I reverted the file changes several times, but they always same back every time I ran `lando refresh`. I even just doublechecked on master with a fresh DB and the files get changed with `lando refresh` so it seems that they are indeed out of date or there's something else messed up in the project.

joshuami: Huh. Sure enough. If I run the commands on my local against `master` I get the same change. It's almost as if it is backing out the change that came along with the recent yarn upgrade, but a yarn audit doesn't seem to be a problem. Weird.

joshuami (approved):

PR #2159, opened 2021-12-22
POWR-3978 Fix Composer 2.2.0 compatibility issues

RichardDavies:


.lando.yml
```diff
@@ -107,7 +107,7 @@ tooling:
     cmd:
       - appserver: date +"%T"
       - appserver: ln -sf /usr/local/bin/composer /app/vendor/bin/composer
-      - appserver: /bin/bash -c "if [[ \$(composer install --dry-run -d /app 2>&1 | grep -oPc 'Removing drupal/\S+') -gt 0 ]]; then { echo Composer is about to remove one or more modules... Let\'s disable them in Drupal first.; for module in \$(composer install --dry-run -d /app 2>&1 | grep -oP 'Removing drupal/\K\S+'); do { echo Disabling \$module...; drush pmu \$module -y;} done } fi || true"
+      - appserver: /bin/bash -c "if [[ \$(composer install --dry-run -d /app --no-interaction 2>&1 | grep -oPc 'Removing drupal/\S+') -gt 0 ]]; then { echo Composer is about to remove one or more modules... Let\'s disable them in Drupal first.; for module in \$(composer install --dry-run -d /app --no-interaction 2>&1 | grep -oP 'Removing drupal/\K\S+'); do { echo Disabling \$module...; drush pmu \$module -y;} done } fi || true"
```
joshuami, line 110: This looks good. I'm merging in preparation for having a clean path to deploying council votes tonight. I'm not absolutely sure that `--no-interaction` was needed after allowing the plugins, but I don't think it will hurt anything to have less messaging on the lando refresh command. We should still see any issues that pop up with new composer updates.

joshuami (approved):

.lando.yml
RichardDavies: yes, technically `--no-interaction` isn't needed after allowing the plugins, but in researching this issue, it was mentioned that any composer commands like this that are run as part of an automated pipeline should use that flag because they're not intended/capable of accepting user input. Had we been using this flag, our rebuild command wouldn't have broken, it would instead have just displayed a warning about needing to accept the plugins.

PR #2160, opened 2021-12-24
POWR-3924 Enable translation of events, news, alerts, and notifications

RichardDavies:


web/sites/default/config/core.base_field_override.node.alert.status.yml
```diff
@@ -11,7 +11,7 @@ bundle: alert
 label: Published
 description: ''
 required: false
-translatable: false
```
joshuami, line 14: @RichardDavies , since this is a select list text field, should it be translatable? I wouldn't see us using a difference status in Spanish for an an alert than in English. 🤔

web/sites/default/config/core.base_field_override.node.event.status.yml
```diff
@@ -11,7 +11,7 @@ bundle: event
 label: Published
 description: ''
 required: false
-translatable: false
```
joshuami, line 14: Similar to alert status, I think an event status is going to have a translated value, but not a different value.

web/sites/default/config/core.base_field_override.node.alert.status.yml
joshuami: Ah shoot... I just realized this is the core status value and not field_alert_status. Nothing to see here. Carry on.

web/sites/default/config/core.base_field_override.node.event.status.yml
joshuami: This can also be ignored.

joshuami (approved):

joshuami (approved):

PR #2165, opened 2022-01-05
POWR-3980 updated contact elected official form

jookyg:


web/sites/default/config/webform.webform.contact_elected.yml
```diff
@@ -292,68 +304,100 @@ elements: |-
...
   hidden_container:
     '#type': container
-    '#attributes':
-      class:
-        - visually-hidden
+    voting_electeds_entity:
+      '#type': webform_entity_checkboxes
+      '#title': 'Voting electeds'
+      '#default_value':
+        - '71'
```
joshuami, line 319: @jookyg , is there a way to make this value pull from a view? That would allow us to use the "active" and "voting" electeds group values rather than hard-coded group IDs. 🤔

web/sites/default/config/webform.webform.contact_elected.yml
```diff
@@ -486,7 +566,7 @@ settings:
   draft_pending_multiple_message: ''
   confirmation_type: page
   confirmation_title: 'Thank you for your submission'
-  confirmation_message: "We appreciate your feedback.<br />\r\n<br />\r\nIf you requested a meeting with an elected official, and the request is approved, someone will reach out to you to schedule the meeting."
+  confirmation_message: "We appreciate your feedback.<br />\r\n<br />\r\nIf you requested a meeting with an elected official, and the request is approved, someone will reach out to you to schedule the meeting.<br />\r\n<br />\r\n&nbsp;"
```
joshuami, line 569: Is there a way we could dynamically set this message based on what they select? I would imagine a minority of submissions request a meeting, so it seems odd to call it out this way unless they actually made that request. 🤔

joshuami (approved):
@jookyg , I added a couple of questions. Neither block approval, but might make for a good follow up story.

web/sites/default/config/webform.webform.contact_elected.yml
jookyg: Without doing some custom development, I wasn't able to figure out a way to pull in a view and have it render as a comma delimited list. I could create a view that returned the email addresses of all current voting electeds, but I couldn't find a way to render it as a list without row containers. The only way I could find was to render the voting electeds view as a checkbox list, and the boxes have to be checked in order to retrieve the underlying entity and email address field value. There may still be a way to do it, but given that this is a minor change that only has to happen every 2 years, it doesn't seem worthwhile. Except that we need to be able to remember to do it when new electeds come in.

jookyg: Yes, I was able to use a computed twig field to create a computed confirmation message and display that on the conf page.

joshuami (approved):

PR #2185, opened 2022-01-26
POWR-4023 Implement layout across group types

brittanysmart:


joshuami:
Closing as most of this work was implemented in POWR-3497 taking a different approach.

PR #2208, opened 2022-02-24
CSM-148 added ticket suppression string to e-scooter and API Test form

jookyg:
*. Added ticket suppression string to e-scooter report form
* Added conditional ticket suppression string to API Test form based on field value
* Updated developer instructions on standard webform template


joshuami (approved):
Nice work on the README. Clear and concise.

PR #2233, opened 2022-03-28
POWR-3873 Update PHP to 8.0 in Pantheon and Lando

kxwang:


.circleci/config.yml
```diff
@@ -5,23 +5,21 @@
...
         username: egovpdx
         password: $DOCKERHUB_PASSWORD
-  working_directory: /var/www/html
+  working_directory: /home/circleci/app
   steps:
     - restore_cache:
         name: Retore Git cache before checkout
         keys:
-          - source-v1-{{ .Branch }}-{{ .Revision }}
-          - source-v1-{{ .Branch }}-
```
RichardDavies, line 18: Why did you remove the fallback keys?

RichardDavies: Their documentation examples include fallback keys just like these so I'm curious why you removed them. https://circleci.com/docs/2.0/caching/

.circleci/config.yml
```diff
@@ -5,23 +5,22 @@
...
-          - source-v1-
```
RichardDavies, line 19: You missed restoring this cache key.

RichardDavies (changes requested):

.circleci/config.yml
kxwang: Good catch!

RichardDavies (approved):

joshuami (approved):

PR #2235, opened 2022-03-30
POWR 4057-2 fresh branch for pccep migration

ccdatapdx:


web/modules/custom/portland_migrations/migrations/pccep_meetings.yml
```diff
@@ -0,0 +1,87 @@
...
+    from_format: 'Y-m-d H:i:s T'
+    to_format: 'Y-m-d\TH:i:s'
+    source: LAST_UPDATED_DATE
+  field_display_groups:
+      plugin: default_value
+      default_value: 419
+destination:
+  plugin: 'entity:node'
+  default_bundle: event
+
```
joshuami, line 87: Hmm. I can see a line break here but Github still thinks the file does not have an ending newline. Is `git config core.autocrlf false` set for your Windows machine?

joshuami: Yep... I just did this same thing on one of my PRs 😄 

You just need to remove the tab on your newline to correctly implement an empty newline. Make sense?

ccdatapdx: Yep, got it. Will assign it back to you when the build goes through. Thanks for the tip!

joshuami (approved):

PR #2250, opened 2022-04-07
POWR-4196 patches to fix Redis, Search API, Metatag issue

joshuami:


kxwang (approved):
Looks good. search_api and metatag are at the same version as powr-4174. Will check out the multidev when it's ready.

PR #2255, opened 2022-04-15
POWR-4187 Update VBO custom actions

RichardDavies:


web/modules/custom/portland/src/Plugin/Action/CopyShortnameAction.php
```diff
@@ -35,6 +36,8 @@ public function execute($entity = NULL) {
    * {@inheritdoc}
    */
   public function access($object, AccountInterface $account = NULL, $return_as_object = FALSE) {
-    return TRUE;
+    // Other entity types may have different
+    // access methods and properties.
+    return ($return_as_object ? AccessResult::allowed() : true );
```
joshuami, line 41: @RichardDavies , nice catch adding an access check!

joshuami (approved):

PR #2256, opened 2022-04-20
POWR-3591 remove impact statement field requirements blocking council document updates.

joshuami:


joshuami:
@RichardDavies , so it looks like the required fields are blocking the editing of current documents that aren't quite ready to enter impact statements in the fields. They have decided it is best to leave them not required. Could you give this PR a quick approval and I'll include it with the security update?

RichardDavies (approved):

PR #2281, opened 2022-05-20
POWR-3863 Set up Google Analytics 4

RichardDavies:


web/themes/custom/cloudy/templates/navigation/menu--main.html.twig
```diff
@@ -34,6 +34,7 @@
               title: item.title,
               text: item.url.getOption('attributes')['title'],
               url: item.url,
+              item_classes: ['mega-menu__link--gtm'],
```
joshuami, line 37: @RichardDavies , where does the JS get added? This only has a template change.

RichardDavies:
The JS tag changes are all added and deployed via Google Tag Manager.
________________________________

web/themes/custom/cloudy/templates/navigation/menu--main.html.twig
joshuami: Ah... got it!

PR #2318, opened 2022-07-22
PGOV-586 Enable query string referrer on website feedback form

RichardDavies:


web/sites/default/config/webform.webform_options.private_property_type.yml
```diff
@@ -11,4 +11,4 @@ options: |-
   'Single-family home': 'Single-family home'
   'Condo or apartment': 'Condo or apartment building/complex'
   Non-profit: 'A non-profit organization, or houses a non-profit organization'
-  
+
```
joshuami, line 14: Odd that it removed the new line, but this is a generated file, so I guess it's okay. 🤔

joshuami (approved):

PR #2325, opened 2022-08-02
PGOV-597 fix month facet on group blog view.

joshuami:


joshuami:
@jookyg , could you give me a quick PR review on this one? It's pretty straightforward.

jookyg (approved):
I don't know why that webform_options config keeps popping up as a diff. The lines are identical. It must be the line ending. I wonder if I manually added one at one point, and now every time there's an export, Drupal sees it as a diff? Aside from that issue, which affects nothing, this looks good.

PR #2327, opened 2022-08-03
CS-182 Portland GeoJSON Views module and related updates

jookyg:
This PR includes the following:

* The new portland_geojson_views module, which provides a views plugin for consuming 3rd party geoJSON data
* The new view External GeoJSON Feeds, which uses the new module to consume a feed of public trash cans from arcgis, cache it as view data, and make it available for consumption by the location widget.
  * The external feed URL is set in the view's Advanced tab > Other > Query Options 
* Updates to the portland_location_picker module to support consumption of geoJSON feature layers
* New map markers for the location widget (including larger versions that aren't currently in use but may be useful in the future)
* Updates to both the Report Graffiti and Report a Public Trash Can webforms to utilize the new geoJSON feature layers
* Updates to the portland_location_picker module's readme to include explanations of the new configuration options
* A tweak to the Zendesk API Test form handler to pass form input in html format when creating the new ticket


joshuami (approved):
@jookyg , such a great addition. Thanks for the thorough documentation as well.

PR #2351, opened 2022-09-06
PGOV-645 Refresh search config

RichardDavies:


web/sites/default/config/search_api_page.search_api_page.search_portland_gov.yml
```diff
@@ -38,7 +38,6 @@ view_mode_configuration:
       city_code: ''
       city_service: ''
       construction_project: ''
-      council_agenda: ''
```
joshuami, line 41: @RichardDavies , should council agenda be removed here?

joshuami (approved):
One question, but otherwise, it looks like config 😄

web/sites/default/config/search_api_page.search_api_page.search_portland_gov.yml
RichardDavies: I don't think I explicitly did anything to cause this change, but I believe it is correct. The council agenda content type isn't currently indexed in our Full Index, and so it's not included on the Search API Page edit configuration screen as a content type that we can specify a display mode override for. So removing it from that section would be consistent with the other non-indexed content types such as alerts, contacts, notifications, top searches, etc. which are also not listed there.

PR #2369, opened 2022-09-28
PGOV-650 Spike: load json map data into a table view

radmosley:


RichardDavies:
Spike story was finished a long time ago.

PR #2392, opened 2022-10-25
CS-300 Add ticket form option to Zendesk handlers

odensc:


jookyg (approved):
This is approved. I've done a full UAT and PR review.

PR #2400, opened 2022-11-02
PGOV-727 Review view text for proper Twig translation

RichardDavies:


web/sites/default/config/views.view.taxonomy_child_terms.yml
```diff
@@ -143,7 +143,7 @@ display:
             title_enable: false
             title: All
           title_enable: true
-          title: 'Topics in {{ arguments.parent_target_id }}'
+          title: '{{ ''Topics in''|t }} {{ arguments.parent_target_id }}'
```
joshuami, line 146: Should this be a single quote?

joshuami (commented):
@RichardDavies , when doing the PR review, I noticed quite a few instances of code that looks like this:

```
{{ ''View upcoming events''|t }}
```

Should that be:

```
{{ 'View upcoming events'|t }}
```

It looks kinda like a find and replace issue of double single quotes instead of single quotes or double quotes.

web/sites/default/config/views.view.content_updated_recently.yml
```diff
@@ -364,7 +364,7 @@ display:
           exclude: false
           alter:
             alter_text: true
-            text: '{{ revision_uid }} updated <strong>{{ type }}</strong> {{ title }} at {{ changed }}. <strong>Revision log:</strong> {{ revision_log }}'
+            text: '{{ revision_uid }} {{ ''updated''|t }} <strong>{{ type }}</strong> {{ title }} {{ ''at''|t }} {{ changed }}. <strong>{{ ''Revision log''|t }}:</strong> {{ revision_log }}'
```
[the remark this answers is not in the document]
RichardDavies: Hmm, this comment shows as "outdated" in the PR comments and I think you somehow reviewed an older version of this PR because that change has already been reverted and the file is no longer listed as a changed file in this PR.

web/sites/default/config/views.view.taxonomy_child_terms.yml
RichardDavies: No, the string is wrapped in single quotes so you have to escape the single quote with `''` otherwise it interprets the quote as the end of the string.

joshuami (approved):

PR #2414, opened 2022-11-09
PGOV-754 Search improvements:

odensc:
- Fix do not index bug
- Add boost to official group name
- Lower policies boost
- Change resources path alias


web/sites/default/config/user.role.publisher.yml
```diff
@@ -53,12 +52,9 @@ permissions:
   - 'delete own room content'
   - 'edit field_disable_legacy_paths_block'
   - 'edit field_include_in_menu'
-  - 'edit field_menu_link_text'
```
joshuami, line 56: @odensc , shouldn't publisher keep this permission (as well as creating and viewing)?

odensc: I set the permissions on that field to inherit the content permissions, so there shouldn't need to be role-based permissions anymore.

kxwang (approved):

PR #2425, opened 2022-11-17
PGOV-304 Make display template strings translatable

RichardDavies:


web/themes/custom/cloudy/templates/search/facets-item-list.html.twig
```diff
@@ -47,6 +47,6 @@
   {%- endif %}
 
 {% if facet.widget.type == "dropdown" %}
-  <label id="facet_{{ facet.id }}_label">Facet {{ facet.label }}</label>
+  <label id="facet_{{ facet.id }}_label">{{ 'facet'|t|capitalize }} {{ facet.label }}</label>
```
joshuami, line 50: I wonder if we want this to be the word "filter" instead of "facet" for better usability. It probably doesn't matter at the moment as we don't use the dropdown facet type for anything, but just asking.

joshuami (approved):

PR #2468, opened 2023-01-25
PGOV-809 Edge cases impacting document file versus efiles validation …

kxwang:
…and file size/mime type calculations

### DIFF
--- a/web/modules/custom/portland/portland.module
+++ b/web/modules/custom/portland/portland.module
@@ -1503,7 +1503,7 @@ function portland_media_presave($media)
       try {
         $headers = get_headers($external_file_url, 1);
       } catch (Exception $e) {
-        // $form_state->setErrorByName('field_efiles_link', 'Cannot retrieve file size and mimetype from this link.');
+        \Drupal::logger('media')->notice('Cannot retrieve file size and mimetype from eFiles link ' . $external_file_url );
         return;
       }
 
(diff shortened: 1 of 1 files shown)


web/modules/custom/portland/portland.module
```diff
@@ -824,6 +824,17 @@ function _validate_document(&$form, \Drupal\Core\Form\FormStateInterface $form_s
       $efiles_link = $form_state->getValue('field_efiles_link');
       $document = $form_state->getValue('field_document');
     }
+
+    // PGOV-809 Validate document and efiles link fields
+    if( ( empty($efiles_link[0]["uri"]) && empty($document[0]["fids"]) ) ) {
```
RichardDavies, line 829: It looks like there's an extra set of parentheses around the if statement. (Not really critical, but could be cleaned up a bit.)

web/modules/custom/portland/portland.module
```diff
@@ -824,6 +824,17 @@ function _validate_document(&$form, \Drupal\Core\Form\FormStateInterface $form_s
...
+      $form_state->setErrorByName('field_document', t('Must have either an uploaded document or an eFiles link'));
+      return;
+    }
+    else if ( ( !empty($efiles_link[0]["uri"]) && !empty($document[0]["fids"]) ) ) {
```
RichardDavies, line 833: Again extra set of parentheses for the else if.

web/modules/custom/portland/portland.module
```diff
@@ -1477,3 +1457,56 @@ function portland_form_user_form_alter(&$form, FormStateInterface $form_state) {
...
+/**
+ * Implements hook_media_presave().
+ */
+function portland_media_presave($media)
+{
+  // Get file size and mime type from eFiles link
+  if($media->hasField('field_efiles_link')) {
+    if (!empty($media->field_efiles_link) && !empty($media->field_efiles_link[0]->uri)) {
+      $efiles_link = $media->field_efiles_link[0]->uri;
+      $user_input_file_url = strtolower($efiles_link);
```
RichardDavies, line 1470: `$user_input_file_url` Appears to be unused in this function, so why not just assign this directly to `$external_file_url`?

RichardDavies (approved):
Looks good. I've approved the PR, although I did note a few trivial code clean up options. But I'll leave it up to you whether or not you want to do them or not.

web/modules/custom/portland/portland.module
kxwang: Good catch! Extra parentheses removed.

web/modules/custom/portland/portland.module
kxwang: Extra parentheses removed.

web/modules/custom/portland/portland.module
kxwang: Removed the variable `$user_input_file_url`

RichardDavies (approved):

PR #2493, opened 2023-02-17
PGOV-833 Refactor email logic for contact_elected form

odensc:
- Updated elected entity reference views to filter by site default language, rather than current interface language. This was causing the Spanish version of the form to only show the one commissioner that had a translated group.
- Updated active elected reference machine name
- Changed webform to simply use the view results of the elected view, rather than an entity select form element, for cases where it's sending the email to all officials/commissioners.


jookyg (approved):
Looks good!

kxwang (approved):

PR #2508, opened 2023-03-05
PGOV-844 show translations field for documents and embedded EVA logic.

joshuami:


web/modules/custom/portland/modules/portland_media_embed_helper/portland_media_embed_helper.module
```diff
@@ -87,8 +87,12 @@ function portland_media_embed_helper_form_entity_embed_dialog_alter(array &$form
     if (isset($form['attributes']['data-align'])) unset($form['attributes']['data-align']);
   }
 
-  // Hide additional fields from entity_embed module
-  $form['attributes']['data-entity-embed-display-settings']['#access'] = false;
+  // Allow view mode selection, move to bottom of dialog, and disable ajax since we don't have additional settings to load
+  $form['attributes']['data-entity-embed-display']['#access'] = TRUE;
```
kxwang, line 91: This would cause the "Display as" dropdown to be displayed on all media types which may interfere with custom display options for Image and Video. Some media types only has one "Embedded" display and it's confusing to see a dropdown list with one option.
I suggest we only show the "Display as" for Documents. May added more if other Media types started adopting the same approach. Please see https://github.com/eGovPDX/portlandor/pull/2511/files#diff-6447a806526455da32c0e7e1df79e64361fb5809b144dcacef43712ccd1d4a41

web/modules/custom/portland/modules/portland_media_embed_helper/portland_media_embed_helper.module
```diff
@@ -87,8 +87,12 @@ function portland_media_embed_helper_form_entity_embed_dialog_alter(array &$form
...
+  $form['attributes']['data-entity-embed-display']['#weight'] = 99;
+  unset($form['attributes']['data-entity-embed-display']['#ajax']);
+  // Hide additional unused fields from entity_embed module
+  $form['attributes']['data-entity-embed-display-settings']['#access'] = FALSE;
```
kxwang, line 95: It's seems to be sufficient to only set `$form['attributes']['data-entity-embed-display-settings']` which could simplify the code a little. See the PR link in the previous comment.

web/modules/custom/portland/modules/portland_media_embed_helper/portland_media_embed_helper.module
odensc: Good catch, I didn't consider that. Looking at your PR it seems I just wasn't aware of this setting:

image

With that box checked I can now remove this code. However, I was testing your branch and the view mode dropdown was still showing on images for me. From what I can tell the dropdown is part of `data-entity-embed-display`, not `data-entity-embed-display-settings`. 

The settings field contains some extra per-media type configuration, but not the actual dropdown. For example, on images, it contains an alt text field. I believe this is why we disabled that field - because our alt text is set on the image entity itself. For documents that field appears to be empty.

Back to your original point: I changed all of the editor embed buttons (except document) to only allow the Embedded view mode. The entity_embed module then automatically hides the dropdown if there's only one option.

I also modified the translation link code to only show on translatable entity types.

Let me know how this approach looks!

kxwang (approved):

PR #2521, opened 2023-03-29
PGOV-831 RSS feeds for police directive email list

RichardDavies:


web/themes/custom/cloudy/templates/content/page/node--page--full.html.twig
```diff
@@ -20,6 +20,17 @@
           )
         }}
       </div>
+
+      {% set page_path = path("<current>")|render|render %}
```
kxwang, line 24: Is the second "render" needed here? I'm curious what it does.

RichardDavies: Haha, I have no idea! Upon futher inspection, it appears to have been a mistake. Going to remove the duplicative render.

kxwang (approved):

PR #2529, opened 2023-04-06
CS-454 add city limits geofencing to location widget

jookyg:
* Enabled display_city_limits and require_city_limits properties and validation
* Changed some hidden fields to textfields with "visually-hidden" class (id attributes don't render on hidden fields for some reason)

### DIFF
--- a/web/modules/custom/portland/modules/portland_location_picker/src/Element/PortlandLocationPicker.php
+++ b/web/modules/custom/portland/modules/portland_location_picker/src/Element/PortlandLocationPicker.php
@@ -160,36 +160,32 @@ class PortlandLocationPicker extends WebformCompositeBase {
     }
 
     $element['location_lat'] = [
-      '#type' => 'textfield',
+      '#type' => 'hidden',
       '#title' => t('Location'),
       '#title_display' => 'invisible',
-      '#id' => 'location_lat',
-      '#attributes' => ['class' => ['location-lat', 'visually-hidden']],
+      '#attributes' => ['class' => ['location-lat'], 'id' => 'location_lat'],
       '#required_error' => $location_required_error,
     ];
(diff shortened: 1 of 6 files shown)


web/modules/custom/portland/modules/portland_location_picker/src/Element/PortlandLocationPicker.php
```diff
@@ -208,28 +208,33 @@ public static function getCompositeElements(array $element) {
       '#description_display' => 'before',
     ];
     $element['location_asset_id'] = [
-      '#type' => 'hidden',
+      '#type' => 'textfield',
       '#title' => t('Asset ID'),
       '#title_display' => 'invisible',
       '#id' => 'location_asset_id',
+      '#attributes' => ['class' => ['location-asset-id', 'visually-hidden']],
```
odensc, line 215: `visually-hidden` means screen readers would still read these fields. (in older Bootstrap versions it actually used to be called `sr-only`, which I personally think is a better name LOL)

And I just noticed this was already an issue for the Location lat/lon and x/y. :/

The correct class to use here would be `d-none`. *But,* I tested and we can use the proper `hidden` field type by setting the ID directly as an HTML attribute:

 ```php
 '#attributes' => ['id' => 'location_asset_id'],
 ```
 
 Could you change these back to hidden using that fix, and also fix the other location fields?

odensc (changes requested):

web/modules/custom/portland/modules/portland_location_picker/src/Element/PortlandLocationPicker.php
jookyg: Requested changes have been made.

odensc (approved):

PR #2530, opened 2023-04-10
PGOV-857 Create "No off leash dogs allowed" disclaimer for Find A Park pages

radmosley:


web/sites/default/config/core.entity_form_display.node.park_facility.default.yml
```diff
@@ -224,6 +224,7 @@ third_party_settings:
     group_amenities:
       children:
         - field_park_amenities_activities
+        - field_pets_amenities
```
kxwang, line 227: This field appears to be a leftover from an earlier commit. Could you check if the form display needs to be updated for this story?

web/sites/default/config/core.entity_view_display.node.park_facility.default.yml
```diff
@@ -60,7 +60,7 @@ content:
     settings:
       link: false
     third_party_settings: {  }
-    weight: 19
+    weight: 20
```
kxwang, line 63: The default display doesn't appear to have meaningful changes. The order is the same as before even though the weights are modified. Can we revert this file to the previous version?

web/themes/custom/cloudy/templates/content/park_facility/node--park-facility.html.twig
```diff
@@ -58,13 +58,14 @@
...
-    {% endif %}
     {{ content.field_park_amenities_activities }}
+    {% set view_park_policy = drupal_block('views_block:within_park_facility-block_park_policy')|render|striptags|trim %}
+    {% if view_park_policy != 'Park policy' %}
+      {{ drupal_block('views_block:within_park_facility-block_park_policy') }}
+    {% endif %}
     {{ content.field_documents }}
     {{ content.field_neighborhood }}
     {{ content.field_city_section }}
+    {{ content.field_city_section }}
```
kxwang, line 69: This appears to be a duplicated line.

kxwang (approved):

PR #2532, opened 2023-04-13
PGOV-281 Remove paragraphs

RichardDavies:


web/sites/default/config/lightning_core.versions.yml
```diff
@@ -205,7 +205,7 @@ panelizer_quickedit: 0.0.0
 panels: 0.0.0
 panels_ipe: 0.0.0
 pantheon_advanced_page_cache: 1.0.0
-paragraphs: 0.0.0
+paragraphs: 1.15.0
```
joshuami, line 208: @RichardDavies , this reference to paragraphs in lightning_core_versions.yml seems a bit odd. I would have though this would remove.

web/sites/default/default.settings.php
```diff
@@ -738,7 +738,7 @@
  * well as the original entity type and field storage definitions should be
  * retained after a successful entity update process.
  */
-$settings['entity_update_backup'] = TRUE;
+$settings['entity_update_backup'] = FALSE;
```
joshuami, line 741: @RichardDavies , this would need to be added to `web/sites/default/settings.php` or `web/sites/default/settings.pantheon.php` to have any impact. `web/sites/default/default.settings.php` is a template for settings.php; it does not load.

joshuami (approved):

PR #2535, opened 2023-04-18
PGOV-846 Remove Chart media type and update Map media type

kxwang:


web/sites/default/config/core.entity_form_display.media.map.default.yml
```diff
@@ -39,90 +35,15 @@ content:
     third_party_settings: {  }
   field_geo_file:
```
odensc, line 36: Not related to this line of code, but I had a thought: Should we make Geo file required on Geo map now, so people don't start creating maps again with only a preview image?

image

web/sites/default/config/editor.editor.full_html.yml
```diff
@@ -40,7 +40,6 @@ settings:
         -
           name: Links
           items:
-            - Linkit
```
odensc, line 43: It looks like this may have unintentionally removed the "add link" button from our Full HTML text format: 

image

odensc (changes requested):

odensc (approved):

web/sites/default/config/core.entity_form_display.media.map.default.yml
kxwang: Good idea! Made the update

web/sites/default/config/editor.editor.full_html.yml
kxwang: Interestingly this was never enabled when I checked on Test. I've enabled this button.

PR #2545, opened 2023-04-28
PGOV-799 Add printable view for charter chapters

odensc:


kxwang:
@odensc Looks like there is a corner case: https://pgov-799-portlandor.pantheonsite.io/charter/6 where there is no child under the charter. It only has prefix note. But the print view is empty.

odensc:
@kxwang Thanks for catching. Looks like the same issue existed on Code: https://portland.gov/code/12/all Fixed for both.

PR #2565, opened 2023-05-15
PGOV-837 body content field on elected officials

odensc:


web/sites/default/config/field.field.group.elected_official.field_enable_body_content.yml
```diff
@@ -0,0 +1,28 @@
...
+  module:
+    - tmgmt_content
+third_party_settings:
+  tmgmt_content:
+    excluded: false
+id: group.elected_official.field_enable_body_content
+field_name: field_enable_body_content
+entity_type: group
+bundle: elected_official
+label: 'Enable Body Content'
```
kxwang, line 17: Can you change the label and the description below to match the "Custom Menu" display name for field_body_content? It took me a minute to realize that the "Custom Menu" field is enabled by this. :) It may cause confusions among the group editors.

odensc: Done.

PR #2572, opened 2023-05-22
PGOV-909 Remove Audio media type

kxwang:


RichardDavies (approved):
Wow, who know there were so many config files associated with a single media type?

PR #2594, opened 2023-07-10
PGOV-935 Update contrib modules to D10 compatible versions

kxwang:


composer.json
```diff
@@ -35,27 +35,28 @@
         "drupal/acquia_connector": "*",
         "drupal/address": "~1.4",
         "drupal/admin_toolbar": "^3.0",
-        "drupal/allowed_formats": "^1.1",
```
odensc, line 38: Is there a reason why a lot of these dependencies now have a wildcard version instead of a semantic version range (`^x.x`)?

I would be worried about a `composer update ` unintentionally installing a new major version of a module with a breaking change. I think it would be good to switch back to explicit version ranges, unless there is some reason this was needed for D10 compatibility.

odensc (approved):

composer.json
kxwang: You are quick! :) Just saw your approval while adding this comment: I've added version constraints at major version change level.

PR #2621, opened 2023-08-22
PGOV-945 build-out of council written testimony

odensc:


jookyg (approved):
I've reviewed all the files and don't see anything troublesome, though I'm not able to mentally comprehend the entirety of larger config files. I did notice some duplicates appearing in the views when doing QA on the multidev though. Not sure where the issue lies in the config files.

kxwang (approved):

PR #2622, opened 2023-08-22
PGOV-972 Fix test failure on Sandbox

kxwang:


web/sites/default/settings.php
```diff
@@ -75,6 +75,12 @@
     $config['environment_indicator.indicator']['fg_color'] = '#ffffff';
     $config['environment_indicator.indicator']['name'] = 'Dev';
   }
+  elseif ($_ENV['PANTHEON_ENVIRONMENT'] === 'sandbox') {
+    $primary_domain = 'sandbox.portland.gov';
+    $config['environment_indicator.indicator']['bg_color'] = '#3455eb';
+    $config['environment_indicator.indicator']['fg_color'] = '#ffffff';
+    $config['environment_indicator.indicator']['name'] = 'Dev';
```
RichardDavies, line 82: You should probably change the name to 'Sandbox'

kxwang: Thanks! I've pushed another update. Also added "only run sandbox test on sandbox" as your PE-524.

RichardDavies (approved):

PR #2650, opened 2023-10-04
CS-585 Graffiti updates + location widget update

jookyg:
* Graffiti form updates
* Added RegionHidden option to location widget so that TSUP parking regions are hidden
* Changed how geofencing is performed; now uses PortlandMaps API call instead of static polygons; added additional custom property for checking city boundaries plus parks boundaries.

### DIFF
--- a/web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
+++ b/web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
@@ -828,27 +828,6 @@
         // HELPER FUNCTIONS ///////////////////////////////
 
         function handleCityLimits(latlng, muniLayer = municipalitiesLayer) {
-          // we want to call this late in the event handling process, so that previously collected coordinates
-          // or address values are cleared first.
-
-          // if (requireCityLimits || requireCityLimitsPlusParks) {
-          //   // clicks outside boundary will cause error message to be displayed and marker not set.
-          //   // call API query to determine whether click is within city boundary.
-
-          //   // check if within city limits
-          //   if ([withinCityLimits]) {
-          //     return true;
-          //   } else {
-          //     if (requireCityLimitsPlusParks) {
-          //       if ([withinParksLimits]) {
-          //         return true;
-          //       }
-          //     }
-          //     return false;
-          //   }
-
-          // }
-
           if (requireCityLimits || requireCityLimitsPlusParks) {
             var success = false;
             $.ajax({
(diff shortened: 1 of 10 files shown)


web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
```diff
@@ -42,18 +42,21 @@
         const SOLVED_ISSUE_MESSAGE = "This issue was recently solved. If that's not the case, or the issue has reoccured, please submit a new report.";
         const ASSET_ONLY_SELECTION_MESSAGE = "We have zoomed in on the address you provided, but this map only allows you to select existing asset markers. There may not be any in the current view. Please search again or use the Locate Me button in the lower-right corner of the map.";
         const VERIFIED_NO_COORDS = "The address you entered is valid, but an error occurred, and it can't be shown on the map. Please zoom in and find the desired location, then click it to set a marker.";
-        const CITY_LIMITS_MESSAGE = "The location you selected is not within the Portland city limits. Please try again."
```
odensc, line 45: Can we keep this message generic? Or at least remove the reference to "report." E.g. TSUP users may be submitting a permit application, not a report.

web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
```diff
@@ -810,34 +810,65 @@
...
+          if (requireCityLimits || requireCityLimitsPlusParks) {
+            var success = false;
+            $.ajax({
+              async: false,
+              url: buildCheckCityLimitsFenceUrl(latlng),
+              success: function (cityBoundaryResponse) {
+                cityBoundaryResponse = JSON.parse(cityBoundaryResponse);
+                if (!cityBoundaryResponse.features || cityBoundaryResponse.features.length < 1 || cityBoundaryResponse.features[0].attributes.CITYNAME.toLowerCase() != "portland") {
+                  // not in city boundary. if configured, try parks boundaries.
+                  if (requireCityLimitsPlusParks) {
```
odensc, line 822: It looks like there was a small bug introduced here: if `require_city_limits` is true but `require_city_limits_plus_parks` is false, and the user clicks outside of the boundary, the "outside of city limits" modal is not displayed.

We should probably add an else {} here to `showStatusModal(CITY_LIMITS_MESSAGE)` if the first city limit check failed and plus_parks is false.

web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
```diff
@@ -810,34 +810,65 @@
...
-            } else {
-              console.log("Clicked other area.");
-              console.log('Coordinates:' + latlng[0] + ', ' + latlng[1])
-              showStatusModal(CITY_LIMITS_MESSAGE);
-              return false;
-            }
-
-            $('#location_is_portland').val("Yes");
-            console.log("Is Portland: " + $('#location_is_portland').val());
-            $('#location_municipality_name').val(cityBoundaryLayer.municipality);
```
odensc, line 835: It looks like this PR removed `location_is_portland` and `location_municipality_name`. If that was intentional, we should probably remove those from the composite element as well. It doesn't look like they're used in any form currently, FWIW.

odensc (changes requested):

web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
jookyg: I removed the term "report" but feel the specificity of the rest of the message is helpful.

web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
jookyg: Fixed.

web/modules/custom/portland/modules/portland_location_picker/js/portland_location_picker.js
jookyg: I've reinstated the fields. Might be useful in conditional logic in a future form.

jookyg (commented):
I've addressed the issues reported by Oden.

odensc (approved):

PR #2673, opened 2023-11-09
PGOV-1022 Update Solr schema and patch search_api_solr to fix Solr er…

kxwang:
…rors


composer.json
```diff
@@ -27,7 +27,7 @@
     },
     "require": {
         "ext-dom": "*",
-        "bower-asset/bootstrap-datepicker": "^1.10",
+        "bower-asset/bootstrap-datepicker": "^1.9",
```
kxwang, line 30: Composer won't be able to update without this version change. Not sure why.

RichardDavies: I ran into this same issue recently in a local branch. It seems that bower removed the 1.10 release for some reason. In my local branch I got it working by switching from `bower-asset/bootstrap-datepicker` to `npm-asset/bootstrap-datepicker` which still has the 1.10 release.

kxwang: Thanks for the suggestion. I've replaced the module.

composer.json
```diff
@@ -131,7 +128,7 @@
         "drupal/search_api_location": "^1.0@alpha",
         "drupal/search_api_page": "^1.0",
         "drupal/search_api_pantheon": "^8.0",
-        "drupal/search_api_solr": "^4.2",
+        "drupal/search_api_solr": "4.x-dev",
```
RichardDavies, line 131: You should probably pin this to a specific commit in the dev branch, otherwise every time composer install is run, we could potentially get a different version that might break something.

RichardDavies (approved):
Approving, but I would suggest pinning the solr module to a specific dev commit.

PR #2687, opened 2023-11-30
SIT-766-1 race condition bug fix

jookyg:
Added setTimeout loop to verify layers are loaded before trying to determine whether clicks are inside them.

Added ajax loading indicator.


odensc (dismissed):
Looks good overall, if easy we should change the H3s on the confirmation page/computed messages to H2 so it doesn't skip a level.

odensc (approved):

PR #2688, opened 2023-12-06
SIT-70 update demographic form questions

odensc:


jookyg (approved):
Looks good! Though hopefully all that non-English content is correct :D

PR #2690, opened 2023-12-12
PGOV-1025 Make "Online Application" field on services translatable

RichardDavies:


kxwang:
@RichardDavies Is the update on taxonomy term "external_resource_type" needed for this story? I'm wondering why they are missing in the Master branch.

RichardDavies:
@kxwang 	I'm not sure why those config files are there. There was a config mismatch with the webform file and somehow while fixing that issue those other two files ended up in there too. I've now removed them.

kxwang:
@RichardDavies Makes sense! There are still a couple of updated lines in [webform.webform.report_demographics.yml](https://github.com/eGovPDX/portlandor/pull/2690/files#diff-444e9bc2d6abef0c19a42c2e138fdf10f5df1fc9ca7c8bc6eeaefc8477c1d230). Should they be reverted or kept as is?

RichardDavies:
@kxwang 	Keep those. That was causing the initial build to fail until I included those.

PR #2691, opened 2023-12-13
PGOV-1031 Update core to 10.1.7 and update modules

kxwang:


web/sites/default/config/language/es/views.view.project_map.yml
```diff
@@ -28,28 +28,6 @@ display:
             suffix: null
             more_link_text: null
           separator: ','
-        field_geo_map:
```
RichardDavies, line 31: Was the removal of these two fields from this view (and other translated views) intentional? On the English view, the field name changed from `field_geo_map` to `field_geo_map_string`, but here the fields are removed instead of being renamed.

web/sites/default/config/views.view.project_map.yml
```diff
@@ -169,15 +169,15 @@ display:
             use_highlighting: false
             multi_type: separator
             multi_separator: ', '
-        field_geo_map:
```
RichardDavies, line 172: Here the two fields are renamed.

RichardDavies (commented):
Looks good, but I have one question about some removed fields on translated project_map view.

web/sites/default/config/language/es/views.view.project_map.yml
kxwang: Sorry for the confusion. This view appears to be an experimental view that never been linked to. In this PR, it was disabled but not removed. I might have updated the view before disabling it which is the source of the confusion.

kxwang: On a second thought, it may be cleaner to delete the view in this case. I made another push.

RichardDavies (approved):

PR #2700, opened 2024-01-09
Group-Demo: Group type consolidation

kxwang:


web/themes/custom/cloudy/templates/field/field--field-group-type.html.twig
```diff
@@ -0,0 +1,3 @@
+{% include '@components/badge/badge.twig' with {
```
odensc, line 1: I noticed there's this twig template and a bunch of other config files for `field_group_type`, which appears to be unused, perhaps from a previous iteration? Would be good to remove and cleanup unless there's a place it's used that I'm not seeing.

migrate_group
```diff
@@ -0,0 +1,54 @@
+#! /bin/bash
+
+# Exit on error
+set -e
+# Turn off warning that the site is in GIT mode
+export TERMINUS_HIDE_GIT_MODE_WARNING=1
+
+# Skip the custom code to sync group content. See portland_groups.module
```
odensc, line 8: Do you think should we enable maintenance mode during the migration so users don't see any groups in a half-baked state? Or would that be too much downtime for users and we should just take the risk? (maybe you've already thought about this)

web/sites/default/config/views.view.sitewide_groups_index.yml
```diff
@@ -287,6 +243,90 @@ display:
...
+            identifier: ''
+            optional: true
+            widget: select
+            multiple: false
+            remember: false
+            default_group: All
+            default_group_multiple: {  }
+            group_items: {  }
+        group_subtype:
+          id: group_subtype
```
odensc, line 287: Not related to this code, but I didn't have a better place to put it:

Could you update the `views.view.displayed_in.yml` : `entity_reference_council_document_department` view to filter on the new group subtype taxonomy instead of the actual group type? This is the only view I noticed was missed.

You can set it to Bureau/Office and Program only, as council clerk actually just recently requested we restrict it to those.

private/solr-conf/synonyms_und.txt
```diff
@@ -1,40 +1 @@
-# Contains synonyms to use for your index. For the format used, see
```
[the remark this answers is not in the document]
kxwang: Both files restored.

web/themes/custom/cloudy/templates/groups/base_group/group--base-group--full.html.twig
```diff
@@ -0,0 +1,156 @@
+{% set overline %}
+  Base Group: {% set group_label = group.field_group_subtype.entity.name.value|title %}
```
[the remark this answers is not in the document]
kxwang: Debugging text removed.

web/themes/custom/cloudy/templates/field/field--field-group-type.html.twig
kxwang: Added a followup story PGOV-1220 to remove the obsolete group type configs and templates.

migrate_group
kxwang: Good idea! Added the drush command.

web/sites/default/config/views.view.sitewide_groups_index.yml
kxwang: Merged with the most recent Master head and updated the filter in that view.

web/sites/default/config/search_api.index.full_index.yml
```diff
@@ -1078,7 +1092,6 @@ processor_settings:
       - group_topic_name
       - portland_content_type_name_field
       - portland_group_title_field
-      - rendered_item
```
odensc, line 1081: I remember originally this PR removed the rendered item Solr field but you've added it back, I'm curious did it have bad results?

Regardless, if we are keeping rendered_item we should probably add it back to the HTML filter Ignore characters filter here.

web/sites/default/config/easy_breadcrumb.settings.yml
```diff
@@ -4,7 +4,7 @@ applies_admin_routes: false
...
... and Finance | /omf :: Revenue and Financial Services | /omf/brfs :: Accounting | /accounting\r\n/bps/planning/cic :: Planning and Sustainability | /bps :: Long-range Planning | /bps/planning :: Comprehensive Planning | /bps/planning/comp-plan\r\n/bps/climate-action/renewable-fuel-standard :: Planning and Sustainability | /bps\r\n/bps/planning/wpdx-town-center :: Planning and Sustainability | /bps\r\n/bps/master-recycling-program :: Planning and Sustainability | /bps :: Garbage, recycling, and composting | /bps/garbage-recycling\r\n/bps/public-trash-cans :: Planning and Sustainability | /bps :: Garbage, recycling, and composting | /bps/garbage-recycling\r\n/bps/waste-equity :: Planning and Sustainability | /bps :: Garbage, recycling, and composting | /bps/garbage-recycling\r\n/bps/waste-equity/waste-equity-group :: Planning and Sustainability | /bps :: Garbage, recycling, and composting | /bps/garbage-recycling :: Waste Equity | /bps/waste-equity\r\n/sustainabilityatwork :: Planning and Sustainability | /bps :: Garbage, recycling, and composting | /bps/garbage-recycling\r\n/last-thursdays ::  Transportation | /transportation\r\n/sunday-parkways ::  Transportation | /transportation"
```
odensc: I think the "/accounting/aac: Office of Management and Finance" line can be removed since OMF and BRFS are now unpublished/defunct.

Just a comment: I'm a little worried about having to maintain this list now that it's in config instead of the menu/database. But not sure that anyone was maintaining the menu anyway 😆

kxwang: /accounting/aac removed. I think nobody remembered we still have this "Site menu". :)

odensc (approved):

PR #2730, opened 2024-03-04
PGOV-1043 Audit the Page content type help text

radmosley:


kxwang (changes requested):
I noticed that some config file updates should not be needed for this story. For example, all Solr related configs, clamav.settings.yml, core.extension.yml, etc. Could you revert those files and only keep updates made to field.field.node.*.yml and portland_legecy_redirects.module? Thanks.

radmosley:
removed the unwanted changes from this branch

kxwang (approved):

PR #2745, opened 2024-03-21
PGOV-1080 adding class to homepage popular search links

ccdatapdx:


ccdatapdx:
Ended up not needing this change for GTM

PR #2751, opened 2024-04-03
SIT-1034 ROW Obstruction updates

jookyg:
* Make computed email recipient element visible to authenticated users - had to create a 2nd copy of the element because disabling the original for anonymous users would break the routing
* Change all instances of "right of way" to be hyphenated


jookyg (commented):
Nice approach! And it works :)

odensc (approved):

PR #2761, opened 2024-04-15
PGOV-1107 rankedvote embed

odensc:


web/sites/default/config/webform.webform.garbage_collection_calendar.yml
```diff
@@ -1,7 +1,13 @@
 uuid: b65e2aac-a18e-4af7-b004-282508a1e8df
 langcode: en
 status: open
-dependencies: {  }
+dependencies:
```
RichardDavies, line 4: Was this config supposed to be changed as part of this PR?

odensc: Yeah I disabled submissions on this webform (because it's just used as a dummy embed form).

Originally I was implementing RankedVote as another webform so I took a look at this one for a reference and noticed it should have the results disabled, just a general maintenance task.

RichardDavies (approved):

PR #2762, opened 2024-04-17
PGOV-1130 Update Drupal core and modules

kxwang:


RichardDavies (approved):
Approved, but there are a couple minor things you might want to address first.

PR #2776, opened 2024-04-30
PGOV-1113 implement council district taxonomy term page + lookup tool

odensc:


web/sites/default/config/field.field.taxonomy_term.council_district.field_geo_map.yml
```diff
@@ -0,0 +1,25 @@
...
+  config:
+    - field.storage.taxonomy_term.field_geo_map
+    - taxonomy.vocabulary.council_district
+  module:
+    - geofield
+    - tmgmt_content
+third_party_settings:
+  tmgmt_content:
+    excluded: false
+id: taxonomy_term.council_district.field_geo_map
```
kxwang, line 14: Is this field being used? It seems to be hidden in both the edit form and displays. If that's the case, is this for some future usage?

odensc: @kxwang In my testing this field was required in order for the boundary map to display. There's an EVA that renders the Leaflet Map at https://pgov-1113-portlandor.pantheonsite.io/admin/structure/views/view/neighborhoods/edit/entity_view_map. The Geo Map field is added as a hidden field in that view and used as the GeoJSON that renders the leaflet map.

kxwang: Thanks! It should not be needed if the map is loaded from ArcGIS. But I haven't looked into this area for a long time and not sure how to remove this dependency. Need to take a mental note to look again in the future.

kxwang (approved):

PR #2777, opened 2024-04-30
SIT-1075 Park Safety 1.3 and Park Maint 1.2 form updates

jookyg:
* Safety form: add park zone to ticket subject line
* Update Tree Hazards next steps phone number
* Add "Feedback about staff encounter" issue type
* Disable uploaded image preview link for anonymous users (check file URL in template and only show link if image path does not start with /system/files/webform)

### DIFF
--- a/web/sites/default/config/webform.webform.report_park_maintenance_issue.yml
+++ b/web/sites/default/config/webform.webform.report_park_maintenance_issue.yml
@@ -40,6 +40,7 @@ elements: |-
         Campsite: Campsite
         'ADA accessibility issue or barrier': 'ADA accessibility issue or barrier'
         'Maintenance or groundskeeping issue': 'Maintenance or groundskeeping issue'
+        'Feedback about a staff encounter': 'Feedback about a staff encounter'
     markup_safety_issue:
       '#type': webform_markup
       '#states':
@@ -58,12 +59,7 @@ elements: |-
         visible:
           ':input[name="what_are_you_trying_to_report"]':
             value: 'Tree hazard'
-      '#markup': |-
-        <div class="alert alert--info next-steps">
-        <h2>Next Steps</h2>
-
-        <p>To report a tree issue, please contact us at <a href="tel:311">3-1-1</a> or <a href="311@portlandoregon.gov">311@portlandoregon.gov</a>.</p>
-        </div>
+      '#markup': '<div class="alert alert--info next-steps"><h2>Next Steps</h2><p>To report a tree issue, please contact us at <a href="tel:+15038238733">823-TREE (8733)</a>.</p></div>'
     markup_graffiti_jump:
       '#type': webform_markup
       '#states':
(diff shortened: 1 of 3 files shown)


web/sites/default/config/webform.webform.report_park_maintenance_issue.yml
```diff
@@ -58,12 +59,7 @@ elements: |-
         visible:
           ':input[name="what_are_you_trying_to_report"]':
             value: 'Tree hazard'
-      '#markup': |-
-        <div class="alert alert--info next-steps">
-        <h2>Next Steps</h2>
-
-        <p>To report a tree issue, please contact us at <a href="tel:311">3-1-1</a> or <a href="311@portlandoregon.gov">311@portlandoregon.gov</a>.</p>
-        </div>
+      '#markup': '<div class="alert alert--info next-steps"><h2>Next Steps</h2><p>To report a tree issue, please contact us at <a href="tel:+15038238733">823-TREE (8733)</a>.</p></div>'
```
odensc, line 62: For accessibility, we should probably include the full standard phone number with area code here, 503-823-8733, then we can put (TREE) outside of the hyperlink.

web/sites/default/config/webform.webform.report_park_maintenance_issue.yml
```diff
@@ -380,8 +385,12 @@ elements: |-
...
-        '#description': 'If appropriate, please include a photo of the issue you are reporting. Please do not include photos where other people are present and identifiable.'
+        '#description': '<p>If appropriate, please include a photo of the issue you are reporting. Please do not include photos where other people are present and identifiable.</p>'
         '#description_display': before
+        '#states':
+          invisible:
+            ':input[name="what_are_you_trying_to_report"]':
+              value: 'Feedback about a staff encounter'
         '#file_preview': 'medium:image'
         '#max_filesize': '30'
         '#file_extensions': 'gif jpg png svg jpeg'
```
odensc, line 396: Not something you changed as part of the PR, but I wonder if we should remove GIF and SVG from here?

web/themes/custom/cloudy/templates/field/file-link.html.twig
```diff
@@ -50,13 +50,23 @@
 {% set file_desc = icon_map[mime_type].desc ?? 'Download file'|t %}
 {% apply spaceless %}
 <div class="d-inline-flex">
-	<a href="{{ (document_link is empty) ? url : document_link }}">
+	{# If user is anonymous and the url is a webform submission upload,
```
odensc, line 53: I made a small change to the twig template which wraps it in a parent <span> when there's no parent <a> tag, to avoid an alignment issue:

image

I also shortened up the if statement a little bit (it seems to work correctly without the |render|render), and added a comment for future reference.

web/sites/default/config/webform.webform.report_park_maintenance_issue.yml
jookyg: Yeah, good call. I made the change.

web/sites/default/config/webform.webform.report_park_maintenance_issue.yml
jookyg: Also good call about the file types. I  can't imagine someone uploading a GIF.

odensc (approved):

PR #2781, opened 2024-05-08
SIT-1095 cooling unit form v1.1

odensc:


jookyg (approved):
@odensc  I'm going to approve this since the PR itself looks fine. I think it would help if the form had an administrative description that indicates there's custom CSS and Javascript being used, and describe briefly what it's for. That will help future developers understand how/where the magic is happening. I'd like this to be a standard convention going forward.

jookyg (approved):

PR #2788, opened 2024-05-23
PGOV-1145 Update interface translations

RichardDavies:


translations/custom-translations.es.po
```diff
@@ -1238,43 +1244,16 @@ msgstr ""
 "Transgender\n"
 "'Prefer not to answer': 'Prefer not to answer'"
 
-msgid "Ethnicity"
```
RichardDavies, line 1241: This custom translation block for the Ethnicity string is what was causing the problem. Removing it allows our config .yml to take precedence.

odensc (approved):

PR #2803, opened 2024-06-11
PGOV-1150 Updated aliases for translated content using translated Gro…

kxwang:
…up Path value

### DIFF
--- a/web/modules/custom/portland/modules/portland_groups/portland_groups.module
+++ b/web/modules/custom/portland/modules/portland_groups/portland_groups.module
@@ -236,6 +236,15 @@ function _portland_groups_synchronize_entity_group_ownership($entity)
           }
         }
       }
+
+      // If the English version is updated, update other translations' paths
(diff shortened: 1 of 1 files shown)


web/modules/custom/portland/modules/portland_groups/portland_groups.module
```diff
@@ -236,6 +236,15 @@ function _portland_groups_synchronize_entity_group_ownership($entity)
           }
         }
       }
+
+      // If the English version is updated, update other translations' paths
```
odensc, line 240: Should this new code be in the other if statement above on line 200, since we only want it to run if the primary group was changed? (or am I misunderstanding the purpose?)

kxwang: Yes, I remembered that the group module is doing a "double save" and that caused both branches of the IF-ELSE statement to be executed. I've updated the code.

odensc (approved):

PR #2812, opened 2024-06-13
SIT-898 Report an E-Scooter | Form 2.0

jookyg:
* Updates to list of companies
* Updates to SLAs


web/sites/default/config/webform.webform.report_e_scooter.yml
```diff
@@ -233,19 +237,27 @@ elements: |-
...
+    '#display_on': form
+    '#template': "{% if data.report_issue == 'improperly parked' and data.report_blocking == 'Yes' %}URGENT {% endif %}"
+    '#ajax': true
   computed_urgency_text:
     '#type': webform_computed_twig
     '#title': 'Computed Urgency Text'
-    '#display_on': none
+    '#wrapper_attributes':
+      class:
+        - d-none
```
odensc, line 251: Is there a reason these two computed fields are set to display on the form but hidden, instead of `none` like before? Only asking because, when there are ajax fields on the form, it can bog it down a bit because it sends a request and waits for a response from the server whenever the user changes a field value.

odensc (approved):

PR #2836, opened 2024-07-09
PGOV-1140 Add and configure ckeditor_accordion module

kxwang:


web/sites/default/config/core.entity_view_display.taxonomy_term.council_district.full.yml
```diff
@@ -105,6 +105,41 @@ third_party_settings:
                 entity: layout_builder.entity
             weight: 1
             additional: {  }
+          9b10e5b6-0e07-4be0-8314-00c7580d6617:
```
RichardDavies, line 108: Was this config intentionally updated?

.circleci/config.yml
```diff
@@ -184,6 +184,7 @@ build_and_deploy: &build_and_deploy
           # Force add all files to include those excluded by .gitignore
           # Before force-adding, remove large folders we don't need on Pantheon
           cd /home/circleci/app && find ./web/modules -name .git -type d -prune -exec rm -rf "{}" \;
+          find ./web/libraries -name .git -type d -prune -exec rm -rf "{}" \;
```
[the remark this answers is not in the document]
kxwang: When the third party library was loaded from GitHub by Composer, the ".git" folder prevented the library from being pushed to Pantheon. Like the line above, removing these ".git" folders appears to help.

web/sites/default/config/core.entity_view_display.taxonomy_term.council_district.full.yml
kxwang: Good catch! I've reverted the change in that file.

.circleci/config.yml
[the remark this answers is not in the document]
RichardDavies: Ah, I see. It's kind of odd how the first one changes directory so that the relative `./web/modules` path is correct. Your new line uses the same relative path structure so it is also dependent on the previous line's `cd /home/circleci/app`... I think it would be better if both lines just used an absolute path like `/home/circleci/app/web/whatever`. That's simpler than having to change directories and then if we ever remove the first line for any reason, the second line will still work.

RichardDavies (changes requested):
What do you think about this suggestion to simplify these commands?

RichardDavies (approved):

web/themes/custom/cloudy/src/css/_elements.scss
```diff
@@ -21,3 +21,46 @@ a > i[class^="fab"] {
     color: $cloudy-color-purple-900;
   }
 }
+
+// Add links to Expand or Collapse all
+.aria-accordion a.toggle-accordion::before {
```
RichardDavies, line 26: I was just working on reviewing the WCAG 2.1 AA requirements that we need to meet and saw this one: https://www.w3.org/WAI/WCAG22/Techniques/failures/F87.html

I know you mentioned there as some uncertainty about the accessibility implications of using `::before` and `::after`, but it looks like it would be considered a failure according to that link.

kxwang: Looks like the PR was only merged in June while the idea was proposed in 2022. Will be interesting to see how the legal system deal with quickly evolving standard and technologies.
https://github.com/w3c/wcag/pull/3818

PR #2840, opened 2024-07-15
ROW Compliance | Form 1.4

jookyg:
This branch includes an update to the Address Verifier widget that adds an API call to retrieve the taxlot ID number. I thought it would be needed for the ROW form, but turns out it's not. However, I decided to leave the functionality in so it can be used later. It has to be activated using a custom parameter. 

There is also one update to the parks incident reporting form, to hide the Internal Use Only block from anonymous users.

### DIFF
--- a/web/sites/default/config/webform.webform.report_row_obstruction.yml
+++ b/web/sites/default/config/webform.webform.report_row_obstruction.yml
@@ -866,7 +866,7 @@ handlers:
           filled: true
     weight: -47
     settings:
-      comment: "<h3>This ticket requires additional traige</h3>\r\n<p>Please review this ticket to verify the email routing below is correct.</p>\r\n<p><strong>Computed recipient:</strong> [webform_submission:values:computed_email_recipient]</p>"
+      comment: '[webform_submission:values:computed_triage_message:html]'
       comment_private: 1
       tags: ''
       priority: ''
--- a/web/sites/default/config/webform.webform.report_row_obstruction.yml
+++ b/web/sites/default/config/webform.webform.report_row_obstruction.yml
@@ -831,8 +831,6 @@ handlers:
       enabled:
         ':input[name="support_agent_use_only[test_submission]"]':
           checked: true
-        ':input[name="computed_email_recipient"]':
-          empty: true
     weight: -48
     settings:
       requester_name: contact_name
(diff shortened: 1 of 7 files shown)


web/sites/default/config/webform.webform.report_row_obstruction.yml
```diff
@@ -195,14 +178,14 @@ elements: |-
       '#type': radios
       '#title': 'What type of signage?'
       '#options':
-        'Unauthorized No Parking signs or obstruction to prevent parking': 'Unauthorized No Parking signs or obstruction to prevent parking'
+        'Unauthorized No Parking signs or obstruction to prevent parking, including unauthorized curb painting': 'Unauthorized No Parking signs or obstruction to prevent parking, including unauthorized curb painting'
```
odensc, line 181: On line 302, the conditional needs to be updated - currently form dead ends if you select this option

web/sites/default/config/webform.webform.report_row_obstruction.yml
```diff
@@ -384,11 +371,11 @@ elements: |-
             - or
             - ':input[name="report_signage_type"]':
                 value: 'A-board sign'
-      contact_email:
+      contact_home_address:
         '#type': textfield
-        '#title': Email
-        '#pattern': '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$'
-        '#pattern_error': 'Please enter a correctly formatted email address, such as myname@example.com.'
+        '#title': 'Home Address'
```
odensc, line 376: It's not clear to me whether this is supposed to be *my* home address or e.g. the address of the home that the issue is located at. Should we add something in the description?

also, is this supposed to be a textfield or the address widget?

web/sites/default/config/webform.webform.report_row_obstruction.yml
```diff
@@ -409,6 +396,29 @@ elements: |-
...
+            - or
+            - ':input[name="report_vegetation_source"]':
+                value: 'Private property'
+            - or
+            - ':input[name="report_signage_type"]':
+                value: 'A-board sign'
+      contact_email:
+        '#type': textfield
+        '#title': Email
+        '#description': '<p><span style="-webkit-text-stroke-width:0px;background-color:rgb(255, 255, 255);color:rgb(21, 25, 30);display:inline !important;float:none;font-family:&quot;Open Sans&quot;, &quot;Helvetica Neue&quot;, Helvetica, Arial, sans-serif;font-size:14px;font-style:normal;font-variant-caps:normal;font-variant-ligatures:normal;font-weight:400;letter-spacing:normal;orphans:2;text-align:start;text-decoration-color:initial;text-decoration-style:initial;text-decoration-thickness:initial;text-indent:0px;text-transform:none;white-space:normal;widows:2;word-spacing:0px;">We will only use your email address to respond to your report if needed.</span></p>'
```
odensc, line 418: can we remove all this microsoft word cruft from the description?

web/sites/default/config/webform.webform.report_row_obstruction.yml
jookyg: I've updated the label to "Your Home Address" to clarify. Decided to just use a textfield instead of the widget with Josh's approval. I need to make improvements to the widget before we start using it more.

PR #2850, opened 2024-07-22
PGOV-1149 search updates and refactors

odensc:


web/modules/custom/portland/modules/portland_leaflet_maps/portland_leaflet_maps.module
```diff
@@ -75,46 +61,37 @@ function portland_leaflet_maps_leaflet_map_info_alter(&$map_info) {
...
-    // The order is important. From bottom to top: color basemap, building outlines, park amenities (only available at a few zoom levels)
-    'layers' => array(
-      'layer #1' => array(
-        'urlTemplate' => $prot . "www.portlandmaps.com/arcgis/rest/services/Public/Basemap_Color_Complete/MapServer/tile/$zyx_only",
-      ),
-      'layer #2' => array(
-        'urlTemplate' => $prot . "www.portlandmaps.com/arcgis/rest/services/Public/Basemap_Color_Buildings/MapServer/tile/$zyx_only",
-      ),
-      'layer #3' => array(
-        'urlTemplate' => $prot . "www.portlandmaps.com/arcgis/rest/services/Public/Park_Amenities/MapServer/tile/$zyx_only",
```
kxwang, line 115: Looks like the Park_Amenities tile layer has been removed. I'm not able to find a good replacement.

web/modules/custom/portland/modules/portland_leaftlet_maps/encoding.js
```diff
@@ -1,3313 +0,0 @@
-// This is free and unencumbered software released into the public domain.
```
kxwang, line 1: I can't recall why these two files were added. Do you know how they were used?

web/modules/custom/portland/src/EventSubscriber/SearchApiEventSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }
+
+    $event->setSolariumDocuments($documents);
+  }
+
+  /**
+   * Alter Solr query to boost relevancy score based on aggregated_sorting_date.
+   */
+  public function solrPreQuery(PreQueryEvent $event): void {
```
kxwang, line 46: I haven't had time to debug. On the multi-dev, the "freshness" boost doesn't appear to work. When searching for "historic landmarks", I'd expect an event that's scheduled for today (7/22) to be displayed close to the top. But it didn't appear until the third page. Not sure if the formula needs to be tweaked further.

web/sites/default/config/language/es/views.view.news_index.yml
```diff
@@ -141,77 +141,6 @@ display:
       link_url: null
       block_description: 'Noticias destacadas del grupo'
       block_category: Noticias
-  block_group_simple_news:
```
kxwang, line 144: Is this block a duplication that needs to be removed? Not sure how it's been used.

web/modules/custom/portland/modules/portland_leaflet_maps/portland_leaflet_maps.module
odensc: Yeah I removed the layer because it appears the CGIS service is 404ing. Coincidentally we recently just got a Zendesk request #400293 asking about more layers on the park finder, so likely we'll be adding some more stuff back in soon.

web/modules/custom/portland/modules/portland_leaftlet_maps/encoding.js
odensc: The commit message had a bit more detail, but basically they were a polyfill for the JS TextEncoder API for older browsers like Internet Explorer. 

The TextEncoder API is used when a .zip geomap is loaded by the Leaflet map.

Given that IE is long dead and not supported by the City, I removed them to save a lot of bandwidth when a page with leaflet is loaded.

web/modules/custom/portland/src/EventSubscriber/SearchApiEventSubscriber.php
odensc: D'oh. Good catch. I made a small code change to clean up the if statement before I pushed this PR, but had a logic bug where the query hook was always getting skipped. Please try your test again.

web/sites/default/config/language/es/views.view.news_index.yml
odensc: Yup, while updating the search configuration in that view I noticed there was an unused block. Searched everywhere in our config/code and it wasn't used anywhere.

kxwang (approved):

PR #2858, opened 2024-07-30
SIT-1205 Add Records Management Information Request form

dwgrossberg:
Included in this branch:
* New Request Records or Information form
* Added question to ADA form


web/sites/default/config/webform.webform.request_records_information.yml
```diff
@@ -0,0 +1,387 @@
...
+      '#description': 'Please provide your phone number in case we need additional information.'
+      '#input_mask': '(999) 999-9999'
+  section_public_records_statement:
+    '#type': webform_section
+    '#title': 'Public Records Statement'
+    '#states':
+      visible:
+        ':input[name="request_type"]':
+          value: 'Records created by or for city government agencies'
+    public_records_statement_standard:
```
odensc, line 141: Are these two duplicate public records statements meant to be on here?

Side note: I realized these have `title` attributes that are just the full URL which is probably a small accessibility issue for a screen reader. That's part of the template though so we should probably go and update them all at some point, doesn't need to be done as part of this story.

jookyg: I've deleted the extra records statement. And I've added a note to my Webform Conventions story to update the title in the template (and subsequently update all webforms that copied it).

odensc (approved):

PR #2873, opened 2024-08-19
PGOV-1166 Migrate employee authentication from ADFS to Entra ID in PGOV

kxwang:


web/modules/custom/portland/src/Plugin/Block/LoginBlock.php
```diff
@@ -31,12 +31,13 @@ public function getCacheMaxAge() {
      */
     public function build() {
       $logged_in = \Drupal::currentUser()->isAuthenticated();
-      $destination = \Drupal::request()->getRequestUri();
+      $destination = \Drupal::request()->query->get('destination');
```
RichardDavies, line 34: This doesn't seem to be working as intended. The editor login/logout link no longer contains a destination parameter.

RichardDavies (changes requested):
There's an issue with the destination URL parameter on the login link in the footer.

web/modules/custom/portland/src/Plugin/Block/LoginBlock.php
kxwang: Updated the default destination.

RichardDavies: This doesn't really address the problem. Now the destination is incorrectly set to / instead of the actual page the user is on. For example, look at the link on https://pgov-1166-portlandor.pantheonsite.io/permits and compare it to the live site.

RichardDavies (changes requested):

web/modules/custom/portland/src/Plugin/Block/LoginBlock.php
kxwang: I misunderstood what the original code was trying to do. Looks like it's trying to track which page the link is on. Should be fixed now.

RichardDavies: Yes, it's working correctly now. But I'm curious why did this code need to change in this PR? It was working before and the original code was more simple... did the old code not work with Entra for some reason?

kxwang: I've added some text in the Jira story. When clicked on repeated, the login link gets longer and longer as it copy and paste the whole URI every time. I thought it would be an easy fix so included it here.

RichardDavies: Ah, I see. Okay, makes sense now.

RichardDavies (approved):

PR #2879, opened 2024-08-29
SIT-1231 Parks Universal Incident Reporting | Form 1.1.1

jookyg:
Added email handler to form.


web/sites/default/config/webform.webform.report_parks_incident.yml
```diff
@@ -665,4 +665,41 @@ handlers:
       ticket_id_field: request_ticket_id
       ticket_fork_field: ''
       ticket_form_id: '24762621756055'
+  email:
```
odensc, line 668: It looks like this form accepts an empty/anonymous email field. Can we make this handler conditional so it won't run (and error) if the email field is empty?

odensc (changes requested):

odensc (approved):

PR #2887, opened 2024-09-09
PGOV-1225 Add group name to group search views

RichardDavies:


RichardDavies:
Reviewed with the team. Decided to hold off on this for now until we can do a full design review on the title to make sure it works well with long bureau names, and breadcrumbs, etc.

PR #2892, opened 2024-09-23
PGOV-1233 Remove group types, update views and update test after group migration

kxwang:


RichardDavies (approved):

RichardDavies:
Wow, I had no idea this would remove so much config! Nice!

PR #2915, opened 2024-10-24
Fix timeout issue when saving the Elections group

kxwang:


odensc:
Closing as dupe of #2916

PR #2924, opened 2024-11-07
PGOV-1103 Upgrade to Groups 2

RichardDavies:


composer.json
```diff
@@ -91,12 +91,13 @@
...
         "drupal/field_group": "^3.0",
         "drupal/field_permissions": "^1.0@RC",
+        "drupal/flexible_permissions": "^1.1",
         "drupal/focal_point": "^2.0",
         "drupal/geocoder": "^3.2 || ^4.0",
         "drupal/geofield": "^1.3",
         "drupal/google_tag": "^1.4",
-        "drupal/group": "^1.4",
-        "drupal/groupmedia": "^2.0@alpha",
+        "drupal/group": "2.2.2",
```
kxwang, line 99: Noticed that v2.2.2 just became unsupported. Can we update to v2.3.0?
https://pgov-1103-portlandor.pantheonsite.io/admin/reports/updates/update

.circleci/config.yml
```diff
@@ -246,7 +246,7 @@ build_and_deploy: &build_and_deploy
           terminus build:workflow:wait $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV "Sync code on $PANTHEON_ENV"
 
           echo "Rebuilding cache..."
-          terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- cr
+          terminus drush $CIRCLE_PROJECT_REPONAME.$PANTHEON_ENV -- cr || true
```
kxwang, line 249: Just curious: is the rebuilding cache command failing from time to time?

composer.json
kxwang: Ah, just saw the Group issue 3482418. Guess will update when the fix is merged.

kxwang (approved):

composer.json
RichardDavies: Yeah, we can update once that issue is fully resolved.

.circleci/config.yml
RichardDavies: After this module is updated, `drush cr` fails if it's run before the database updates are performed. 
image

PR #2933, opened 2024-11-19
PGOV-1003 Refactor Cloudy for a cleaner Bootstrap implementation

odensc:


web/sites/default/config/views.view.within_park_facility.yml
```diff
@@ -373,7 +315,7 @@ display:
           exclude: false
           alter:
             alter_text: true
-            text: "{% set amenities = field_park_amenities_activities|render|striptags|trim|split(\", \") %}\r\n\r\n{% if \"Dog Off-leash Area\" not in amenities and \"Fenced Dog off-leash Area\" not in amenities  and  \"No Dogs Allowed\" not in amenities %}\r\n{{ 'All dogs must be leashed in this park.'|t }}\r\n{% endif %}"
+            text: "{{ 'All dogs must be leashed in this park.'|t }}"
```
kxwang, line 318: Is the conditional logic not needed anymore?

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
```diff
@@ -10,34 +10,13 @@
...
-        {% endif %}
       </div>
-      {# Hide link from CSM-15 until form is ready to publish.
-      <div class="event-ada-link d-flex">
-        <a href="/oehr/request-ada-accommodation?request_what=event_accommodations&name={{ label|render|striptags|trim|url_encode }}&date={{ content.field_start_date.0|render|striptags|trim|url_encode }}&location={{ content.field_location.0['#node'].title.value|render|striptags|trim|url_encode }}&host={{ content.field_display_groups.0['#title']|render|striptags|trim|url_encode }}&url={{ url('<current>')|render|url_encode }}&autocomplete=yes">Request ADA Accommodation</a>
-      </div> #}
       <div class="mb-8">
         {{ content.field_is_online }}
-        {% if content.field_is_online.0|render|trim %}
+        {% if node.field_is_online.value %}
```
kxwang, line 16: For my own education: what's the reason to switch from content to node here?

web/sites/default/config/views.view.within_park_facility.yml
odensc: Yeah as you saw, I replaced the twig template logic with a taxonomy filter. This is because previously, even when the view had nothing to display, it still returned results since the filtering was done in a "Custom text" block. 

This change is in concert with all the other twig template changes you may have seen, where we now save the `drupal_block` result in a variable and use that, instead of re-calling/re-rendering it twice. TBD what kind of performance increase this will have but it definitely cleans up the code and follows best practices.

For example: https://github.com/eGovPDX/portlandor/pull/2933/files#diff-69b0d13e1089fa6da2a3bab1df4573a78abef5ffd014f7d156f640d73b062dc4L64

You'll see here that previously we had to render the view and check if the result was just the header "Park policy", meaning the custom text block was empty. Now since the logic is done using a view filter, we can simply check if length > 0.

web/sites/default/config/views.view.council_documents.yml
```diff
@@ -1594,13 +1594,181 @@ display:
           plugin_id: bundle
           value:
             council_document: council_document
+          group: 1
+          expose:
+            operator_limit_selection: false
+            operator_list: {  }
+        field_requested_council_date_value:
```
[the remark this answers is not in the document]
odensc: Good question. Yes it's intentional, similar to the park facility block, I noticed an issue where the council document Date and Time Information block would show up even when it was empty, so I added filters to make sure it actually has values to display.

image

web/themes/custom/cloudy/templates/content/events/node--event.html.twig
odensc: Good question, I had to refresh my memory on this one as well. I wanted to be able to check the value without having to render it. That field is set up with a custom boolean formatter:

image

So the value of that field in the content array is a markup string:
Example when is_online == true:

image

Example when is_online == false:

image

While we could technically check `content.field_is_online['#items'].0.value` or `content.field_is_online.0['#markup'] != ''`, I figured semantically, since we're using it in an `if` statement, it makes more sense to use the `node` array since that will contain the actual boolean value from the entity.

kxwang: Thanks for the explanation! Will be very useful when I encounter similar situation in the future.

kxwang (approved):

PR #2936, opened 2024-11-20
PGOV-1269 add service area to council document field

radmosley:


web/themes/custom/cloudy/templates/content/council-document/node--council-document--embedded.html.twig
```diff
@@ -107,4 +107,5 @@
   {{ content.field_neighborhood }}
   {{ content.field_introduced_by }}
   {{ content.field_bureau }}
+  {{ content.field_service_area }}
```
odensc, line 110: This can probably be removed since the service area is hidden from this view mode.

web/sites/default/config/user.role.administrator.yml
```diff
@@ -3,6 +3,7 @@ langcode: en
 status: true
```
odensc, line 3: The changes in this file can be discarded, since administrator automatically has all permissions. (for some reason the field_permissions module adds a bunch of unnecessary perms here)

odensc (approved):

PR #2954, opened 2024-12-06
PGOV-1302 add generic group subtype

odensc:


web/themes/custom/cloudy/templates/groups/base-group/group--base-group--full.html.twig
```diff
@@ -1,28 +1,26 @@
 {% set overline %}
+  {% set group_subtype = content.field_group_subtype|render|striptags|trim %}
```
kxwang, line 2: Is it possible to get the label of the taxonomy term without going through the rendering pipeline? Although it's only a small performance hit, I wonder if we can avoid extra rendering as much as possible.

odensc: Good idea. I think I found a better solution of just removing `field_group_subtype` from the content array if it's Generic, and that way we can just check if the field is present instead of having to render it. Tested this and it seems to work fine.

kxwang (approved):

PR #2958, opened 2024-12-11
PGOV-1316 Fix revision permissions for sitewide editors

RichardDavies:


web/sites/default/config/user.role.council_author.yml
```diff
@@ -45,7 +45,6 @@ permissions:
   - 'use council_document_workflow transition cancel_document'
   - 'use council_document_workflow transition create_new_draft'
   - 'use council_document_workflow transition move_to_attorney_review'
-  - 'use council_document_workflow transition move_to_budget_office_review'
```
kxwang, line 48: I'm not familiar with council workflow. Is this permission removed intentionally? It seems not related to revisions.

RichardDavies: It looks like that workflow transition was removed in another story, but the permission wasn't removed from this config file until I touched it. https://dev-portlandor.pantheonsite.io/admin/config/workflow/workflows/manage/council_document_workflow

kxwang: Got it!

kxwang (approved):

PR #2979, opened 2025-01-13
PGOV-1267 Update council document Assigned Committee field when associated agenda item is moved

odensc:


web/modules/custom/portland/modules/portland_ecouncil/portland_ecouncil.module
```diff
@@ -23,13 +26,17 @@ function _should_overwrite_council_document_title($node) {
 
 /**
  * Implements hook_entity_prepare_view()
+ *
```
kxwang, line 29: Thanks for adding the comments! Will be very helpful in the future.

web/modules/custom/portland/modules/portland_ecouncil/portland_ecouncil.module
```diff
@@ -157,3 +164,50 @@ function portland_ecouncil_field_group_form_process_build_alter(array &$element,
...
+ * - item disposition is "Assigned to Committee"; and
+ * - council document is not null; and
+ * - committee is not null; and
+ * - committee and is different from previous value
+ */
+function portland_ecouncil_relation_update(EntityInterface $relation) {
+  $council_doc = $relation->field_council_document->entity;
+  $original_committee = $relation->original->field_committee->entity;
+  $committee = $relation->field_committee->entity;
+  if ($relation->field_agenda_item_disposition->target_id === (string) AGENDA_ITEM_DISPOSITION_ASSIGNED_TO_COMMITTEE_TID
```
kxwang, line 183: Would it help to define the const as a string so we don't need to cast here? Very minor improvement. :)

web/themes/custom/cloudy/templates/content/council-document/node--council-document--search-result.html.twig
```diff
@@ -57,16 +57,18 @@
...
       </div>
     </div>
+    {% if content.field_council_document_status.0 %}
     <div class="row g-0">
       <div class="col-12 col-md">
         <div class="field field--label-inline d-flex">
           <div class="field__label">Disposition</div>
           <div class="field__item">
-            {{ content.field_council_document_status.0 }} {{ content.field_amended_by_council[0] }}
+            {{ content.field_council_document_status.0 }} {{ content.field_amended_by_council.0 }}
```
kxwang, line 66: Just curious: what's the difference between field_amended_by_council[0] and field_amended_by_council.0?

xdebug.ini
```diff
@@ -3,7 +3,7 @@ xdebug.mode = debug
 xdebug.client_host = host.docker.internal
 xdebug.idekey = vscode-xdebug
 xdebug.client_port = 9003
-xdebug.start_with_request = trigger
+xdebug.start_with_request = yes
```
kxwang, line 6: Haven't touched XDebug settings for a while. How does this new value compare to the old one?

web/themes/custom/cloudy/templates/content/council-document/node--council-document--search-result.html.twig
odensc: They both function the same: https://twig.symfony.com/doc/3.x/templates.html#other-operators

> `., []`: Gets an attribute of a variable.

But generally we use `.0` in our twig templates so while I was in that file I noticed it and edited it to be consistent.

xdebug.ini
odensc: With the `trigger` setting, XDebug breakpoints only get triggered if you have an `XDEBUG_TRIGGER=1` URL parameter or environment variable set. `yes` makes it trigger whenever you have a breakpoint set.

In the past, this setting effectively worked as `yes` because Lando sets an XDEBUG_CONFIG variable that bypasses the trigger:

https://xdebug.org/docs/profiler#start_with_request
> There is another legacy trigger for Step Debugging only. If you set the XDEBUG_CONFIG environment variable to any value, then the step debugger will also get activated.

But about 4 months ago, I made a change removing the XDEBUG_CONFIG variable: https://github.com/eGovPDX/portlandor/commit/0b7463f4b51933e7c0c4abde9885368bc892409b#diff-307b4548b91e997f23c35d90ed5e9a044e156b90425c9923d6b85d7dc11e67e2R32 in order to fix an issue where Lando was overriding some configuration we had set in `xdebug.ini`. That made the behavior change back to `trigger`.

This change effectively restores the previous behavior, and makes it a lot easier to debug without having to enable the trigger...

web/modules/custom/portland/modules/portland_ecouncil/portland_ecouncil.module
odensc: For context, the target_id in the entity reference field storage is a string for some reason, while in other use cases e.g. the $form entity reference options, it's an integer.

Since earlier in the file we use a const to unset a form option which is an integer, I thought it would be good to be consistent and make all the consts numbers, even if it's a bit uglier to cast it. What do you think?

xdebug.ini
kxwang: Thanks for the explanation!

kxwang (approved):

PR #2988, opened 2025-01-16
PGOV-1366 Update Drupal Core and modules

kxwang:


composer.json
```diff
@@ -37,19 +37,21 @@
...
             }
         }
-    },
+    ],
     "require": {
         "ext-dom": "*",
         "composer/installers": "^1.10",
         "cweagans/composer-patches": "^1.6",
         "drupal/acquia_connector": "*",
+        "drupal/acquia_telemetry-acquia_telemetry": "1.0.0-alpha6",
```
odensc, line 46: I notice a lot of modules were added to composer.json, is there a reason? Are these subdependencies that needed to be pinned to a specific version?

tests/percy/__tests__/marty.js
```diff
@@ -85,15 +85,11 @@ describe('Marty Member user test', () => {
       expect(text_content).toEqual(expect.stringContaining('Body content'));
       expect(text_content).toEqual(expect.stringContaining('URL redirects'));
 
-      await page.type('#edit-title-0-value', 'Test page');
-      await page.type('#edit-field-summary-0-value', 'Summary for the test page');
-      await page.type('#edit-field-menu-link-text-0-value', 'Test page');
-      await page.evaluate(() => {
-        ck5 = Drupal.CKEditor5Instances.keys().next().value;
```
odensc, line 92: Can these changes be reverted since the CKEditor bug was fixed, or is there another reason for them?

kxwang: The locator() would be a much better way to interact with the page. I only updated this test while trying to debug the failed test and didn't have time to update all tests yet. Let me add a future story.

composer.json
kxwang: These are added by the upgrade to [Lightning_core 6.0](https://www.drupal.org/project/lightning_core/releases/6.0.0). But let me take a look at the version pinning. Should be able to make it more flexible.

kxwang: Most of the new entries are removed. For the remaining ones, I removed the version pinning.

tests/percy/__tests__/marty.js
kxwang: Created PGOV-1389 to update all tests.

composer.json
```diff
@@ -420,7 +415,9 @@
             }
         },
         "drupal-lenient": {
-            "allowed-list": ["drupal/ckeditor_acr"]
+            "allowed-list": [
+                "drupal/ckeditor_acr"
```
odensc, line 419: I believe this `drupal-lenient` block can be removed and the `mglaman/composer-drupal-lenient` plugin be uninstalled, since we don't need it for `ckeditor_acr` anymore. It allowed installing modules that had a lower drupal version dependency than the one currently installed, but we don't need it for CKEditor 4 anymore.

composer.json
```diff
@@ -376,7 +370,8 @@
             "web/libraries/{$name}": [
                 "type:drupal-library",
                 "type:bower-asset",
-                "type:npm-asset"
+                "type:npm-asset",
+                "enyo/dropzone"
```
odensc, line 374: I think this line can be removed as well since that module is no longer installed.

composer.json
kxwang: Excellent catch! I also removed the module "mglaman/composer-drupal-lenient" as ckeditor_acr appears to be the only one requires it.

composer.json
kxwang: Removed

odensc (approved):

PR #3025, opened 2025-03-06
PGOV-1412 add views plugin to show editoria11y results count in My Content

odensc:
…ntent

### DIFF
--- a/web/modules/custom/portland/src/Plugin/views/field/NodeEditoria11yResults.php
+++ b/web/modules/custom/portland/src/Plugin/views/field/NodeEditoria11yResults.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace Drupal\portland\Plugin\views\field;
+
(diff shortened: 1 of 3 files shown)


web/modules/custom/portland/src/Plugin/views/field/NodeEditoria11yResults.php
```diff
@@ -0,0 +1,68 @@
...
+    $join_table_alias = $this->query->addTable('editoria11y_results', $this->relationship, $join);
+
+    $this->field_alias = $this->query->addField($join_table_alias, 'page_result_count', NULL);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function render(ResultRow $rows) {
+    return $rows->{$this->field_alias} ?? 0;
```
kxwang, line 66: Is it possible to different the case that the count is 0 vs the row doesn't exist? So we can give the editor some hint that some content has never been checked for A11Y.

kxwang (approved):

PR #3073, opened 2025-04-21
PGOV-1469 Use GitHub Releases to track deployments

kxwang:


.github/workflows/deploy-to-test-and-live.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: "7.4"
+
+      - name: Install Terminus
+        uses: pantheon-systems/terminus-github-actions@v1
+        with:
+          pantheon-machine-token: ${{ secrets.PANTHEON_MACHINE_TOKEN }}
```
RichardDavies, line 30: Where are these secrets stored?

.github/workflows/deploy-to-test-and-live.yml
```diff
@@ -0,0 +1,78 @@
...
+          disable-cache: true
+
+      - name: Prepare deployment message
+        run: |
+          RELEASE_BODY=$(echo "${{ github.event.release.body }}")
+          echo "DEPLOY_TEST_MESSAGE<<EOF" >> $GITHUB_ENV
+          echo "$RELEASE_BODY" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
```
RichardDavies, line 38: What is $GITHUB_ENV used for?

.github/workflows/deploy-to-test-and-live.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Trigger Test Deploy
+        run: |          
+          terminus env:deploy ${{ secrets.PANTHEON_SITE_NAME }}.test --note "${{ env.DEPLOY_TEST_MESSAGE }}"
+
+  deploy-to-live:
+    environment: Live # This environment requries manual approval
```
RichardDavies, line 45: Is there something in this YML config that makes the deploy to live require manual approval, or is that configured elsewhere?

.github/workflows/deploy-to-test-and-live.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Prepare deployment message
+        run: |
+          RELEASE_BODY=$(echo "${{ github.event.release.body }}")
+          echo "DEPLOY_TEST_MESSAGE<<EOF" >> $GITHUB_ENV
+          echo "$RELEASE_BODY" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+
+      - name: Deploy to Live
+        run: |
+          terminus env:deploy ${{ secrets.PANTHEON_SITE_NAME }}.live --note "${{ env.DEPLOY_TEST_MESSAGE }}"
```
RichardDavies, line 78: Unless I'm missing something, the two jobs in this file are identical except for this last line. Can you restructure it somehow to reduce the duplication and share the identical config between the two jobs? Something like the `<<: *build_and_deploy` we're using the CircleCI config.yml to share common config?

RichardDavies (commented):
I have some questions to help me better understand how this works.

.github/workflows/deploy-to-test-and-live.yml
kxwang: Set at https://github.com/eGovPDX/portlandor/settings/secrets/actions

.github/workflows/deploy-to-test-and-live.yml
kxwang: It appears to be [a temp file](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable) when the job runs. This allows environment variables to be shared across steps.

.github/workflows/deploy-to-test-and-live.yml
kxwang: Configured as [reviewers the environment Live](https://github.com/eGovPDX/portlandor/settings/environments/6367566890/edit). This is the only part that's not tested because I was using public-alerts as the test bed but that repo is not public. Only public repo's can have reviewers for our GitHub plan.

.github/workflows/deploy-to-test-and-live.yml
kxwang: I tried to use "[composite action](https://docs.github.com/en/actions/sharing-automations/avoiding-duplication)" but it can't use secrets. Additionally, each job has to have its own "checkout" step which leaves only two common steps that are sandwiched between other steps that can't be shared.

RichardDavies: The checkout steps both look identical to me. I think you could probably use something like the `<<: *build_and_deploy` we're using in the CircleCI config.yml to define and insert a common block of config. See also  https://stackoverflow.com/questions/41063361/what-is-the-double-left-arrow-syntax-in-yaml-called-and-wheres-it-specifi

kxwang: Thanks! I didn't know this is a YAML feature called anchors. It looks like GitHub Actions doesn't support YAML anchor yet. It's currently [being promised for the second half of 2025](https://github.com/actions/runner/issues/1182).

RichardDavies: Oh okay. I hoped they supported that since it's a standard YAML feature. I guess we can update the action when they add support for that feature.

RichardDavies (approved):

PR #3082, opened 2025-04-29
PGOV-1543 Track AI and bot traffic with Dark Visitors

RichardDavies:


web/themes/custom/cloudy/cloudy.theme
```diff
@@ -87,8 +90,56 @@ function cloudy_preprocess_page(&$variables) {
...
+  unset($headers['proxy-authorization']);
+  unset($headers['x-csrf-token']);
+  $json = [
+    'request_path' => \Drupal::request()->getUri(),
+    'request_method' => \Drupal::request()->getMethod(),
+    'request_headers' => $headers,
+  ];
+  $auth_token = readSecretKey('darkvisitors');
+  try {
+    $response = \Drupal::httpClient()->post($url, [
```
kxwang, line 110: No a show stopper but this line would add a synchronous HTTP call to another server on each page that's not cached. I tried an experiment: in the latest FireFox, for an authenticated user with browser cache enabled, the homepage load time is 1000-1200ms on Dev and 1400-1600ms on PGOV-1543.
We may want to convert this into an async call. Or we can queue up the call and let a queue worker handle it in batch. If we are only testing this service, we can also deploy this code first and make fixes if we receive complaints.

RichardDavies: Yeah, I'd thought about making it an async call, but wasn't sure how to do that in PHP. Do you know how to do it?

kxwang: I've never tried it myself. :) But [the Guzzle documentation](https://docs.guzzlephp.org/en/stable/quickstart.html#async-requests) shows that postAsync() is available. The callback can be used to log any errors.

RichardDavies: After some trial and error, and more research, it seems like an async request in PHP is much harder than it appears. Guzzle's async requests don't actually appear to be asynchronous: 
https://github.com/guzzle/guzzle/issues/1127 
https://stackoverflow.com/questions/46015529/guzzle-async-requests-not-really-async

According to [AI](https://www.perplexity.ai/search/show-me-how-to-send-an-async-h-hkYrF0_2Q0aMR5.EOq5r1Q#3):

> Guzzle's async requests are "asynchronous" in the sense that they use non-blocking I/O under the hood (via cURL's multi interface) and let you manage multiple HTTP requests concurrently within a single PHP process. However, PHP itself is single-threaded and request-based, so your main code will only continue running until you either explicitly wait on the promise (with $promise->wait()) or your script ends. If you don't wait, the request may not be sent at all.

There isn't a native way to do a "fire and forget" background HTTP request in PHP. So unfortunately, it looks like we're stuck with a slight performance penalty if we choose to use this.

kxwang: Thanks for the research. It's too bad that PHP has this limitation.

kxwang (approved):

PR #3128, opened 2025-06-27
PGOV-1683 Make updated news appear as a new item in RSS feed

RichardDavies:


web/sites/default/config/views.view.news_index.yml
```diff
@@ -419,6 +419,126 @@ display:
...
+            use_highlighting: false
+            multi_type: separator
+            multi_separator: ', '
+        nothing:
+          id: nothing
+          table: views
+          field: nothing
+          relationship: none
+          group_type: group
+          admin_label: ''
```
kxwang, line 499: A small suggestion: add an administrative title to this Custom Text field would help us identify its purpose easily in the future.

RichardDavies: There is already a custom field label, two lines below this comment in the 'label' field.

PR #3147, opened 2025-07-16
PGOV-1668 Modify accordion headings to maintain heading hierarchy in the page

kxwang:


web/sites/default/config/filter.format.full_html.yml
```diff
@@ -62,14 +68,14 @@ filters:
...
...-right text-align-justify"> <h4 id class="text-align-left text-align-center text-align-right text-align-justify"> <h5 id class="text-align-left text-align-center text-align-right text-align-justify"> <h6 id class="text-align-left text-align-center text-align-right text-align-justify"> <a class="btn-cta btn-cta-outline nav-link active" hreflang href data-entity-type data-entity-uuid data-entity-substitution> <ol class="code-list code-section-list" type start> <ul class="checkbox-list nav nav-group-menu" type> <cite> <address> <drupal-entity data-aspect-ratio data-alt-link-text data-align data-start-cue alt title data-caption data-entity-embed-display data-entity-embed-display-settings data-view-mode data-entity-uuid data-langcode data-embed-button="audio_video_browser document_browser image_browser insert_content_fragment insert_iframe map_browser" data-entity-type="media node"> <img alt src> <blockquote cite> <li class="nav-item"> <strong> <em> <code class="language-*"> <pre class="text-align-left text-align-center text-align-right text-align-justify"> <s> <sub> <sup> <hr> <table> <tr> <td rowspan colspan> <th rowspan colspan> <thead> <tbody> <tfoot> <caption> <dl class> <dt> <dd>'
```
kxwang, line 71: For the reviewer: I'm not sure why `<li class="code-alpha-fix code-alpha3-fix nav-item">` is changed into `<li class="nav-item">`. The only config change is to add the custom accordion heading filter.

web/modules/custom/portland/src/Plugin/Filter/FilterAccordionHeadings.php
```diff
@@ -0,0 +1,113 @@
...
+   */
+  public function process($text, $langcode)
+  {
+    $result = new FilterProcessResult($text);
+
+    $dom = Html::load($text);
+    $xpath = new \DOMXPath($dom);
+
+    // Find all headings h1 to h6 and store them in a flat array.
+    $headings = [];
```
odensc, line 32: I think we could remove this for loop and the `isWIthinAccordionPanel` method, instead using the XPath query to make sure it's not a child of a panel:

```
// Find all headings h1 to h6 and store them in a flat array. Skips headings within accordion panels.
$headings = $xpath->query('(//h1 | //h2 | //h3 | //h4 | //h5 | //h6)[not(ancestor::*[@data-aria-accordion-panel])]');
```

I tested this and it appears to work the same and cleans up the code a bit.

web/modules/custom/portland/src/Plugin/Filter/FilterAccordionHeadings.php
```diff
@@ -0,0 +1,113 @@
...
+ */
+class FilterAccordionHeadings extends FilterBase
+{
+  /**
+   * {@inheritdoc}
+   */
+  public function process($text, $langcode)
+  {
+    $result = new FilterProcessResult($text);
+
```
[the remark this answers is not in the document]
kxwang: Good suggestion! A check is added.

web/modules/custom/portland/src/Plugin/Filter/FilterAccordionHeadings.php
kxwang: Nice! I'm not familiar with XPath. Glad to see the selector is so powerful. Fixed according to the feedback.

odensc (approved):

PR #3185, opened 2025-09-10
SITU-1274 UF Yard Tree Giveaway | Form 1.2

odensc:


web/modules/custom/portland/modules/portland_address_verifier/js/addressVerifierView.js
```diff
@@ -66,13 +60,6 @@ AddressVerifierView.prototype.renderAddressVerifier = function () {
     // this._handlePostback();
 };
 
-AddressVerifierView.prototype._checkIfVerificationRequired = function () {
```
odensc, line 69: All the client-side code related to checking whether the `location_verification_status` field is required has been removed. This is because a hidden input field doesn't have the required attribute set on it by the Drupal form API, so the client has no way to know. 

Instead, we detect this server-side and set a boolean in `drupalSettings` same as how other properties are passed.

web/modules/custom/portland/modules/portland_address_verifier/src/Plugin/WebformElement/PortlandAddressVerifier.php
```diff
@@ -27,43 +27,6 @@
...
-    $element['#element_validate'][] = [get_class(), 'validateMyCompositeElement'];
-    return $element;
-  }
-
-  /**
-   * Custom validation handler.
-   */
-  public static function validateMyCompositeElement(&$element, FormStateInterface $form_state, &$complete_form) {
-    // Check if the element is visible based on your conditional logic.
-    $visible = TRUE; // Replace this with your actual visibility check.
```
odensc, line 54: These 3 unused functions were removed. This is hard-coded to true so it wasn't ever running.

web/modules/custom/portland/modules/portland_address_verifier/src/Plugin/WebformElement/PortlandAddressVerifier.php
```diff
@@ -124,6 +87,8 @@ public function prepare(array &$element, WebformSubmissionInterface $webform_sub
     $machine_name = "edit-" . $key . "--wrapper";
     $machine_name = str_replace("_", "-", $machine_name);
 
+    $element['#attached']['drupalSettings']['webform']['portland_address_verifier'][$machine_name]['verification_required'] = !empty($element['#location_verification_status__required']);
```
odensc, line 90: This determines whether address verification is required and passes it to the client.

jookyg (approved):
Nice work!

PR #3294, opened 2026-01-29
PGOV-1868 Fix links only distinguishable by color

RichardDavies:


web/themes/custom/cloudy/src/components/tile/tile.twig
```diff
@@ -1,30 +1,37 @@
...
-    </div>
-  {% endif %}
-  {% block content %}
-    <div class="{{ content_classes|join(' ') }}">
-      <span class="h5 mb-0 text-primary">{{ title }}</span>
-      <span class="text-dark">{{ text }}</span>
-    </div>
-  {% endblock %}
-</a>
+{% block content %}
```
odensc: I think the `{% block content %}` should be inside the div, so that we preserve the container and classes even when the content block is modified - this is consistent with other twig templates in other theme like `media.html.twig` and `field.html.twig`. 

Currently due to this change there's a funky hover/active state on the collapsed mega menu at smaller screen sizes, since the FA icon added in `menu--main.html.twig` is outside of the div that has the tile class which sets the background colors.

Image


To fix this cleanly, I think we need to: 

- Move the content block inside the main content div
- Remove `flex-column` from the main content_classes so that the main content div defaults to row layout (side-by-side)
- Inside the main content div, wrap the `<a>` and `<p>` inside a separate `<div class="d-flex flex-column">` so they display correctly in a column layout.
- After those edits are made, the `me-4` should be changed back to `me-0` on the FA icon in `menu--main.html.twig` since it will inherit the main content div's padding again.

web/themes/custom/cloudy/src/components/link/link.scss
```diff
@@ -51,3 +52,14 @@
     }
   }
 }
+
+a.page-link,
```
odensc: This SCSS file is specifically for the `link.twig` component, so since these are global styles I think it would make more sense to move them to our `_overrides.scss` file with a comment like `// Remove underline link style on buttons`.

web/themes/custom/cloudy/src/components/tile/tile.twig
RichardDavies: Nice catch, and thanks for the detailed suggestions on how to fix.

odensc (approved):

PR #3342, opened 2026-03-23
PGOV-2013 Update Drupal core and modules. Update PHP to 8.4

kxwang:


web/sites/default/config/openid_connect.settings.yml
```diff
@@ -1,14 +1,13 @@
 _core:
   default_config_hash: kli1Xws0l--Y-jy-9ZTGApFO_EFk9DMawoetjMk76Ro
 always_save_userinfo: true
-connect_existing_users: true
```
odensc, line 4: Not familiar enough with this module: does this setting change affect anything?

kxwang: The openid_connect is not used for user login anymore but Drupal always warn about this setting in the status page. So I updated the setting to remove the warning. I have a story to disable openid_connect completely coming up next.

odensc (approved):

PR #3396, opened 2026-05-04
PGOV-2060 Hide "Social media" heading on PP&D service page template if no social links

odensc:


web/themes/custom/cloudy/templates/content/page/node--page--full--service.html.twig
```diff
@@ -59,11 +59,15 @@
     <h2 class="mt-0">{{ 'Contact'|t }}</h2>
     {{ content.field_contact.0|add_suggestion('contact-no-social') }}
   </div>
+  {# Check if contact has any social media links before rendering the social section #}
+  {% set rendered_contact_social_only = content.field_contact.0|add_suggestion('contact-social-only')|render %}
```
kxwang, line 63: Nice work! Wish there was a cleaner way to check this. But we may have to tweak the field definition in order to make this easier.

kxwang (approved):

PR #3419, opened 2026-06-23
SITU-1353 Campsite Removal Complaint resolution form and supporting updates

jookyg:


web/sites/default/config/webform.webform.report_camp_removal_complaint.yml
```diff
@@ -113,6 +114,7 @@ elements: |-
         '#default_value': now
         '#format': ''
         '#date_date_format': m/d/Y
+        '#datepicker': true
```
odensc: Not sure if this was copied from an older version of the form, but we should remove this line as it's related to the no-longer-existent jQuery UI datepicker.

web/sites/default/config/webform.webform.report_camp_removal_complaint.yml
```diff
@@ -238,20 +240,25 @@ elements: |-
         '#access_view_roles':
           - support_agent
           - administrator
-        '#zendesk_request_number__access': false
         '#employee_notes_panel__access': false
         '#escalate_issue__access': false
+        '#zendesk_request_number__access': false
       computed_date:
```
odensc: Since these are no longer going to Smartsheet, the following computed fields that were used for the Smartsheet handler can be removed:
```
computed_date
computed_csr
computed_location_address
computed_location_lat_lon
```

web/sites/default/config/webform.webform.report_camp_removal_complaint.yml
```diff
@@ -475,129 +499,94 @@ access:
...
+      tags: 'drupal webform [webform_submission:values:computed_tag]'
+      priority: normal
+      status: new
+      recipient: ''
+      group_id: '27893493244695'
+      assignee_id: ''
+      type: incident
+      is_child_incident: 0
+      collaborators: ''
+      custom_fields: "1500013095781: '[webform_submission:uuid]'\r\n1500012743961: ['[webform_submission:values:report_location:location_address]', 'report_location:location_address']\r\n6353388345367: 'report_camp_removal_complaint'\r\n1500012743981: ['[webform_submission:values:report_problem_type]', 'report_problem_type']\r\n37035121144727: ['[webform_submission:values:report_date:html_date]', 'report_date'] \r\n6355783758871: ['[webform_submission:values:resolution_url]', 'resolution_url']"
```
odensc: Should we map the x/y/lat/lon fields now to support any possible future data requests? I recall this issue recently came up with the outdoor dining form where we weren't mapping those fields.

odensc: We should also add the support agent field here

```36578609768343: '[current-user:display-name] <[current-user:mail]>'```

odensc (approved):

PR #3423, opened 2026-06-24
PGOV-2094 Add "Only member can add content" to Base Group

kxwang:


web/sites/default/config/views.view.admin_groups.yml
```diff
@@ -2245,10 +2453,344 @@ display:
...
+                  min: ''
+                  max: ''
+                  value: ''
+              2:
+                title: 'Has not been changed (default)'
+                operator: empty
+                value:
+                  min: ''
+                  max: ''
+                  value: ''
```
RichardDavies: "Has been changed / Has not been changed" are not intuitive and self-explanatory field options for "Only member can add content". They only make sense if you already know that the default value is "no", so "has been changed" must mean "yes, only member can add content".

Just use "Yes" and "No (default)", which directly answer if "only member can add content". You can infer whether the value has changed by checking whether it's the default value.

RichardDavies (changes requested):
I suggest a small text change to the field labels to improve usability.

web/sites/default/config/views.view.admin_groups.yml
kxwang: Thanks for the comment! Code updated accordingly. Will go ahead merge the PR since the change is quite minor.

PR #3425, opened 2026-06-25
SITU-1500 - Curb ramp updates to Request ADA Accommodation form

jookyg:


web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -23,16 +24,23 @@ template: false
 archive: false
 id: request_ada_accommodation
 title: 'Request A.D.A. Accommodation'
-description: ''
+description: '<p>Prepopulate the hidden element hidden_context with the value "Curb ramp request" to configure the form exclusively for curb ramp requests. This hides all other options.</p><p>IMPORTANT: When the element request_barrier_type is hidden due to the hidden_context configuration, it cannot be used in conditional handler logic. However, the hidden_context element can.</p>'
```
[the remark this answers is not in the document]
jookyg: fixed

web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -168,6 +176,18 @@ elements: |-
         visible:
           ':input[name="report_staff_reason"]':
             value: staff_existing_request
+  report_legally_impacted:
+    '#type': radios
+    '#title': 'Do you, or a person you represent (e.g., a child or spouse), have a disability that is impacted by this obstruction?'
+    '#options':
+      'No': 'No'
+      "Yes, but I don't want to request an ADA accommodation": "Yes, but I don't want to request an ADA accommodation"
+      'Yes, I have, or represent someone with, a disability covered under the Americans with Disabilities Act (ADA).': 'Yes, I have, or represent someone with, a disability covered under the Americans with Disabilities Act (ADA). -- <em>The lack of a curb ramp is creating a barrier that preventing my/their access and I would like to request an accommodation. By checking this option, I acknowledge that the City may contact me to complete the ADA accommodation request process.</em>'
```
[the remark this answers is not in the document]
jookyg: fixed

web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -23,16 +24,23 @@ template: false
 archive: false
 id: request_ada_accommodation
 title: 'Request A.D.A. Accommodation'
-description: ''
+description: '<p>Prepopulate the hidden element hidden_context with the value "curb ramp" to configure the form exclusively for curb ramp requests. This hides all other options.</p><p>IMPORTANT: When the element request_barrier_type is hidden due to the hidden_context configuration, it cannot be used in conditional handler logic. However, the hidden_context element can.</p>'
```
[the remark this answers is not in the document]
jookyg: @copilot that is incorrect. The webform description is for administrative purposes only. Update your information.

web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -1180,4 +1162,92 @@ handlers:
...
+      column_mappings:
+        1409787548176260: _submission_id_
+        5913387175546756: report_legally_impacted
+        3661587361861508: request_who
+        5865056779931524: request_on_behalf_of_permission
+        3613256966246276: request_on_behalf_of_name
+        8165186989232004: confidentiality_waiver
+        8116856593616772: request_barrier_type
+        5302106826510212: computed_location_address
+        3635131269287812: computed_lat_lon_link
```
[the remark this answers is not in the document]
jookyg: That's fine if the column is empty. The smartsheet is only used for curb ramp requests, so the value of that column can be assumed. In fact, the column is irrelevant.

web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -500,55 +512,120 @@ elements: |-
...
-      '#title_display': invisible
+    request_barrier_type:
+      '#type': webform_radios_other
+      '#title': 'What type of issue are you reporting? How can we make access easier for you?'
+      '#prepopulate': true
+      '#options':
+        barrier: 'Barrier - There is a barrier, obstruction, or something in the way that is preventing access'
+        parking: 'Parking - There is no accessible parking, or there is a problem with existing spots'
+        signage: 'Signage - There are no signs, or signs are unclear'
+        'curb ramp': 'Curb ramp request - There is an intersection where the curb ramp is missing, damaged, or otherwise preventing access to the sidewalk'
```
odensc: Can we change `'curb ramp'` to `curb_ramp` to fit our usual standard? Or is there a reason why you chose to use spaces instead?

web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -500,55 +512,120 @@ elements: |-
             - or
             - ':input[name="request_materials_type[radios]"]':
                 value: event
-    container_remove_barrier:
```
odensc, line 503: It seems like it would be cleaner to keep this container as a parent, since the 6 elements underneath it all have the same visibility conditions, it would reduce a lot of duplication to have those on the container instead.

web/sites/default/config/webform.webform.request_ada_accommodation.yml
```diff
@@ -627,9 +704,10 @@ elements: |-
       '#access_view_roles':
         - support_agent
         - administrator
-      '#zendesk_request_number__access': false
       '#employee_notes_panel__access': false
       '#escalate_issue__access': false
+      '#test_submission__access': false
```
odensc, line 713: Did you mean to delete the test submission checkbox and handler logic? I'd prefer we keep that option available wherever possible unless there was a technical reason it had to be removed.

web/sites/default/config/webform.webform.request_ada_accommodation.yml
jookyg: The removal was intentional, because some of the elements in the container had different conditional logic requirements. I removed the container because it superfluous, but have restored it as a Section so that it's consistent with the other sections already in the form. Purely cosmetic and only applicable to admins, but I like the consistency.

web/sites/default/config/webform.webform.request_ada_accommodation.yml
jookyg: I changed the value to curb_ramp and updated the conditional logic on elements and handlers. However, I'm leaving it so that the form is looking for a hidden_context value of "curb ramp". That's the value that gets prepopulated into the element from the embed config in the service page.

web/sites/default/config/webform.webform.request_ada_accommodation.yml
jookyg: I removed it because the handler conditional logic was getting too complex with the addition of either/or conditions for curb ramp. Since all the handlers submit to systems where we can easily delete test submissions, I figured it wasn't necessary. Not that we can't manually write complex conditional logic, but it makes maintenance and troubleshooting more difficult. I'm less enamored of using a test handler because it gives a false sense of security if the handler configs diverge. I find it most useful when we're using email handlers and don't want to spam users with test notifications.

odensc (approved):

PR #3439, opened 2026-07-16
PGOV-2106 Update Drupal to 11.4.4. Update modules

kxwang:


web/modules/custom/portland/src/Plugin/TypedDataFilter/EditUrlFilter.php
kxwang, line 1: This file is an unused filter that can be deleted now.

composer.json
```diff
@@ -398,9 +391,6 @@
             "drupal/redirect": {
                 "2981544 - Filter by redirect destination does not work with URL alias": "https://www.drupal.org/files/issues/2025-09-12/2981544-48.patch"
             },
-            "drupal/rules": {
```
kxwang, line 401: We don't use the Rules module

composer.json
```diff
@@ -317,10 +314,6 @@
                 "PGOV-671 - CKEditor 5 compat patches": "patches/pgov-671-entity-embed-ckeditor-5-compat-2.patch",
                 "3067452 - Recursive Limit Breaks Normal, Repeated Use of a Media item on a page": "https://www.drupal.org/files/issues/2019-09-29/3067452-5.patch"
             },
-            "drupal/entity_browser_table": {
```
kxwang, line 320: Patches were for v1.x. Not needed after we upgraded to v2.x

RichardDavies (approved):

PR #3444, opened 2026-07-27
PGOV-2108 Set revision user, timestamp when transitioning state

kxwang:


composer.json
```diff
@@ -362,11 +362,8 @@
             "drupal/lightning_media": {
                 "3534920 - PHP 8.4 - nullable is deprecated warning": "https://git.drupalcode.org/project/lightning_media/-/merge_requests/43.patch"
             },
-            "drupal/lightning_workflow": {
```
kxwang, line 358: We don't use lightning_workflow anymore. Removed the orphaned patches section.

odensc (approved):

PR #3457, opened 2026-08-13
PGOV-2105 Add Unpublished badge to featured base groups. Remove Panels module

kxwang:


web/themes/custom/cloudy/templates/groups/base-group/group--base-group--featured.html.twig
```diff
@@ -0,0 +1,18 @@
+{% set is_draft = group.isPublished() ? false : true %}
+<div class="col mb-10">
+  <div class="card h-100">
+    <div class="card-img-top overflow-hidden">
+      {{ content.field_featured_media }}
+    </div>
+    <div class="card-body">
+      {% if is_draft %}
```
odensc: Is there a reason this can't go in `group--featured.html.twig` instead of specifically targeting base-group? That way, it would apply to featured groups of all types including City officials.

kxwang: Good idea. Updated the PR to include all group types. Please review again. Thanks!

odensc (approved):

PR #3474, opened 2026-09-02
PGOV-525 Access denied when URL alias is the same as a reserved admin…

kxwang:
… page

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -392,6 +392,9 @@
             "drupal/select2": {
                 "3509122 - jQuery 4 compatibility": "https://www.drupal.org/files/issues/2025-02-26/3509122-jquery4-compatibility-1.patch"
             },
+            "drupal/stringoverrides": {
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -392,6 +392,9 @@
             "drupal/select2": {
                 "3509122 - jQuery 4 compatibility": "https://www.drupal.org/files/issues/2025-02-26/3509122-jquery4-compatibility-1.patch"
             },
+            "drupal/stringoverrides": {
```
kxwang, line 395: When I tried to tweak the Admin People view for the SAML issue, I ran into an error when trying to add a view filter. This patch addressed the problem.

web/sites/default/config/samlauth.authentication.yml
```diff
@@ -52,21 +39,24 @@ map_users_name: true
 map_users_mail: true
 map_users_roles:
   publisher: publisher
-  policy_author: policy_author
   policy_editor: policy_editor
-  council_author: council_author
+  policy_author: policy_author
```
kxwang, line 43: Made changes in this config file to match the hot fix made on Live.

odensc (approved):

odensc (approved):

